PR #11, opened 2020-03-23
Development/comrpc tests

pwielders:
Added testing parameters to the OutOfProcess Test plugin


pwielders:
And a fix for the WebKit to always first Suspend the plugin (if not done by Thunder commands) and then stop it.

PR #15, opened 2020-03-25
Development/IOConnector

pwielders:


IOConnector/Reporter.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "Handler.h"
+#include "TimedInput.h"
+
+namespace WPEFramework {
+
+namespace Plugin {
+
+    class Reporter : public IHandler {
+    private:
+        Reporter();
```
MFransen69: no =delete because of recent changes? If so below override is still used

PR #16, opened 2020-03-25
[IOConnector] Fix handler's Start/Stop integer size

rjpdasilva:
All internal variables for dealing with the handler's 'Start' and 'Stop'
settings are 'uint32_t', but the associated members in the respective
config class are 'Core::JSON::DecUInt8', which results in undesired
values (wrap around, truncated) for values larger than 255 (MAX_UINT8).

This way, use 'Core::JSON::DecUInt32' instead on the config class.

### DIFF
--- a/IOConnector/IOConnector.h
+++ b/IOConnector/IOConnector.h
@@ -172,8 +172,8 @@ namespace Plugin {
                 public:
                     Core::JSON::String Name;
                     Core::JSON::String Config;
-                    Core::JSON::DecUInt8 Start;
-                    Core::JSON::DecUInt8 End;
+                    Core::JSON::DecUInt32 Start;
+                    Core::JSON::DecUInt32 End;
                 };
 
                 enum mode {


pwielders:
The start/stop time is in seconds on the configuration. Internally the plugin works with milliSeconds. Since we need to multiply by 1000 I do not want to raise the suggestion that lagers values than 2^32 / 1000 work hence why DecUInt8 should be more than enough. If required (waiting times > 255 seconds needed, we could scale it up to a DecUInt16 but I think there is no real use case for.

rjpdasilva:
Shall I rework it to scale to DecUInt16 instead or do you want to just drop this?
Dropping is fine by me, 255 seconds does seem enough and you're right about the internal types are to store milliseconds, so no real problem for me in keeping it with DecUInt8, but I can also easily change this to DecUInt16 if that makes sense.
Let me know.

pwielders:
I am fine with both. I do think that 255 seconds on a pin-input is more than enough but we also thought that a uint8_t for the pin numbers was more than enough seems that there are boards out there that have pinnumber 507. So I think in the future it might be that someone says I want to use this feature to trigger only if a pin is kept low for more than an hour 60*60 =3600, that would still fit a UINt16 and does not harm the concept. SO I have a slight prefernce to scale it up to uint16_t for the future but if you think it is not worth the while, lets drop it :-)
I leave it up to you.

rjpdasilva:
Will open a new version of this PR with the suggested changes.

PR #20, opened 2020-03-26
added playbackStopped call in OCDMi server side

goruklu:
Needs https://github.com/rdkcentral/Thunder/pull/35 merged first
Signed-off-by: Gurdal Oruklu <gurdal_oruklu@comcast.com>


goruklu:
@MFransen69 I discussed this with Pierre and agreed to rename the function to make it clear what it does.

PR #23, opened 2020-03-27
Added initial support for WiFi auto-connect after boot.

neeraj-deshpande:


pwielders:
As discussed on skype, I think it good to add the reconnect functionality based on this added flag.

WifiControl/WifiControl.config
```diff
@@ -1,2 +1,6 @@
 set (autostart true)
 set (preconditions Platform)
+map()
+    kv(autoconnect false)
```
pwielders: Let by default make the autoconnect true. If there is a config stored in the persisten location, lets make the Wifi plugin in charge of it connecting and reconnecting. If it has to be under JavaScript control, it can beconfigured to "false" but if for example the Javascript that does this can only be reached over the Wifi it is better to automatically connect (AND reconenct) on lost of connection.

neeraj-deshpande:
As discussed on Skype, following changes are done:
Modified WifiControl plugin to create stateful implementation of auto-connect job.
Modified to set autoconnect variable in WifiControl.json to true as default.
Modified auto-connect job to connect to preferred SSID first.
Modified auto-connect job to sort available networks as per signal strength and attempt auto-connection in the same order.
Modified to notify CTRL_EVENT_SCAN_RESULTS event.
Created asynchronous implementation of connect as synchronous connect will block workpool thread for long time.
Modified to notify OK response as event.
Modified to handle invalid PSK condition.
Added RECONNECTING & AUTHENTICATION states to auto connect job.

PR #31, opened 2020-04-01
[Power]Adding Amlogic implementation changes

arun-madhavan-013:
- **CMakeList changes:** Abstracted source files, headers and related libraries using WPE_POWER_PLATFORM_IMPL switch which is added in meta-wpe/recipes-wpe/wpeframework/include/power.inc file.

- Added **Amlogic** implementation of Power Plugin.


Power/PowerImplementation/Amlogic/PowerImplementation.cpp
```diff
@@ -0,0 +1,256 @@
...
+				fd = open(sEarlySuspendTriggerFile, O_WRONLY);
+				if (fd > 0) {
+					write(fd, "1", strlen("1"));
+					close(fd);
+				} else {
+					TRACE(Trace::Error, (_T("Not able to access %s."),
+								sEarlySuspendTriggerFile));
+				}
+
+				fd = open(sStateFile, O_WRONLY);
```
pwielders, line 153: As you are using file descriptors, i think there is no need to make a busy-waiting loop. File descriptos can be monitored using the Coe::esouceMonitor. Whenever there is something to read on these desciptors, you get a notification. This way you can avoind ceating a thead (8MB stack by default) and a continous loop of checking CPU cycle waist.
A good example can be found here to use FD: https://github.com/rdkcentral/ThunderNanoServices/blob/master/IOConnector/GPIO.h#L35

arun-madhavan-013: Thanks for the reference. Will re-write accordingly.

arun-madhavan-013: Is Coe::resouceMonitor in GPIO implementation similar to poll concept on file descriptors?
But that only works when we wait for data to be read from that FD and process I assume.
Here I'm not sure how can I re-frame the implementation using Coe::resouceMonitor.

In Amlogic platform; only STR power saving mode is supported.

The worker thread implementation only waits for the User Input(Set state API [and Power Key also?]).
If set state JSONRPC API requests STR mode with a timeout TO, these changes are updated in state variables(_newState & _timeout); triggers an event(_wait.SetEvent()) and the API returns.
Worker waiting at _wait.lock() gets activated and does the job; wait till user set TO, writes to /sys/power/state (this is a blocking write).
This triggers device to enter STR mode.
After hardware power ON; device resumes from STR and resets the configurations for normal mode.

If Worker implementation is not there; the JSON RPC set state request will never return as it will remain in blocking state until device resumes from STR.

Could you please explain on your point.

pwielders: Is there a specificiation how the Power mode changes work?
I see you are readig some files, my assumptoion is that if you get a trigger to change powermode, you first signal that in a call, than you wait for data to become available on a filedescriptor (That would be the ResourceMonitor) if you receive the data, you are or, on that state or need to issue another command.
In both situations I assume you take some actiona and again wait fro a signal on the FD.
Waiting for changes on FD (files, sockets) can be done with the ResourceMonitor (epoll under the hood) and it will call you if something happens, you can than take a short action <10ms) to do something.

arun-madhavan-013: I'm not clear about specifications.
There is no read happening. Only wait for User trigger(JSON RPC call) is added in the implementation section.
Please see [this](https://github.com/arun-madhavan-013/ThunderNanoServices/blob/AmlogicPowerPluginIntegration/Power/PowerImplementation/Amlogic/AmlogicPowerPlugin.md).

arun-madhavan-013: In Amlogic case; only two modes are there; ON or STA.
If user is able to issue trigger and plugin receives; device is in ON state.
The data being received is already available in the JSON RPC payload received at [here](https://github.com/arun-madhavan-013/ThunderNanoServices/blob/1725876615d44bf19b6bd23d323c2ace64867fce/Power/PowerImplementation/Amlogic/PowerImplementation.cpp#L230). There is no more data to be received in this plugin's case. Then this function triggers a Core::Event _wait on which Worker is [pending with a lock](https://github.com/arun-madhavan-013/ThunderNanoServices/blob/1725876615d44bf19b6bd23d323c2ace64867fce/Power/PowerImplementation/Amlogic/PowerImplementation.cpp#L120).

arun-madhavan-013: Without the thread I'm not sure how the PowerImplementation::SetState() API call will return... say the JSON RPC received with {state:mem, timeout:50}....will it wait to elapse 50sec.
Or the JSON RPC caller is not expecting the response back..?

pwielders:
I see that you copied the Broadcom implementation to make the AMLogic implementation.
I think the Broadcom implementation was not the bet to copy. I have refactored the broadcom to avoid the thread usage as well there (temoved it)
Also, while going revieuwing your implementation in moe detail, I lened that it is not an Amlogic specific imlementation but a Linux geeric solution, using kernel features.
So I suggest to not call he implementation AMLogic but just Linux.  I created an implementation of that functionality with the refactored Power Plugin on this branch.
I did not yet compile the Linux implementation, I did compile the Broadcom one.
Maybe you can have a look, test and extend this (linux) implementation. The documentation link I follwed from the linux kernel information is on the op of the implementation.
https://github.com/rdkcentral/ThunderNanoServices/tree/development/PowerPlugin

arun-madhavan-013:
> I see that you copied the Broadcom implementation to make the AMLogic implementation.
> I think the Broadcom implementation was not the bet to copy. I have refactored the broadcom to avoid the thread usage as well there (temoved it)
> Also, while going revieuwing your implementation in moe detail, I lened that it is not an Amlogic specific imlementation but a Linux geeric solution, using kernel features.
> So I suggest to not call he implementation AMLogic but just Linux. I created an implementation of that functionality with the refactored Power Plugin on this branch.
> I did not yet compile the Linux implementation, I did compile the Broadcom one.
> Maybe you can have a look, test and extend this (linux) implementation. The documentation link I follwed from the linux kernel information is on the op of the implementation.
> https://github.com/rdkcentral/ThunderNanoServices/tree/development/PowerPlugin

Thanks @pwielders , will expand and submit a new PR after validation.
Abandoning this PR since its no longer needed.

PR #36, opened 2020-04-07
[PowerPlugin] Updated linux implementation with validation.

arun-madhavan-013:
Change require https://github.com/WebPlatformForEmbedded/meta-wpe/pull/428 also for selecting the implementation from recipe.


arun-madhavan-013:
Adding reference of respective IPower changes in Thunder.
https://github.com/rdkcentral/Thunder/tree/development/PowerPlugin

Power/PowerImplementation/Linux/PowerImplementation.cpp
```diff
@@ -122,61 +161,65 @@ class PowerImplementation {
...
-
-            /* We will be able to write state only if we are in 'On' State. */
-            ::write(_stateFile, newMode->label, newMode->length);
-
-            ::write(_triggerFile, "0", 1);
-
+            uint8_t value = -1;
+            TRACE(Trace::Information, (_T("Scheduled Job: change Power mode to %d after %dsec."),
+                        state, waitTimeOut));
+            SleepMs(waitTimeOut * 1000);
```
pwielders, line 195: This line can be removed, as the Job is now scheduled to be fired after this waitTime is passed.

Power/PowerImplementation/Linux/PowerImplementation.cpp
```diff
@@ -122,61 +161,65 @@ class PowerImplementation {
 
             if (job.IsValid() == false) {
                 result = false;
-            }
-            else {
-                static_cast<Job&>(_job).Set(state);
+            } else {
+                static_cast<Job&>(_job).Set(state, waitTimeOut);
```
pwielders, line 165: Check here if it is a new state and if it is a new state schedule the job at Now().Add(waitTImeout * 1000); This way the thread is not busy waiting an can be used untill the timeout occures. Also no need to keep the time than in the job, as the Job will only be executed after the number of seconds have passed.
Just remove the line indicated below, and call here:
Core::IWorkerPool::Instance().Schedule(Core::Time.Now().Add(waitTimeOut * 1000, job);

Power/PowerImplementation/Linux/PowerImplementation.cpp
```diff
@@ -122,61 +161,65 @@ class PowerImplementation {
...
-            /* Kept for future extension. */
-            ::write(_triggerFile, "1", 1);
-
-            /* We will be able to write state only if we are in 'On' State. */
-            ::write(_stateFile, newMode->label, newMode->length);
-
-            ::write(_triggerFile, "0", 1);
-
+            uint8_t value = -1;
+            TRACE(Trace::Information, (_T("Scheduled Job: change Power mode to %d after %dsec."),
```
pwielders, line 193: If you want to see the TRACE line, move it then to the remark avbove and TRACE when you Schedule..

Power/PowerImplementation/Linux/PowerImplementation.cpp
```diff
@@ -190,13 +233,17 @@ class PowerImplementation {
...
+    uint32_t _supportedModes;
+
 private:
-    PCState _currentState;
     int _stateFile;
     int _triggerFile;
     power_state_change _callback;
     void* _userData;
     Core::ThreadPool::JobType<Job> _job;
+    Core::CriticalSection _adminLock;
```
pwielders, line 245: I think the _adminLock is not used anymore, so just remove it.

pwielders:
@woutermeek  Iis the merge done correctly with the master ? I see a lot of ork already merged also in here ? Is that expected ?

Power/PowerImplementation/Linux/PowerImplementation.cpp
arun-madhavan-013: SleepMs() is removed now since Core::IWorkerPool::Instance().Schedule() does the same purpose as suggested.

PR #45, opened 2020-04-17
NetworkControl: avoid IP request from the dispatcher loop, since it ending with deadlock

HaseenaSainul:
If start wlan0 in Dynamic mode with out a AP connected, this request will get failed and re request for IP using RequestFailed(). That will try to clear dispatcher using StopWatchdog(). But in this case this call will happened from the Dispatcher function only. Hence its creating a deadlock case.

### DIFF
--- a/NetworkControl/NetworkControl.cpp
+++ b/NetworkControl/NetworkControl.cpp
@@ -521,8 +521,8 @@ namespace Plugin
             } else {
                 std::map<const string, Core::ProxyType<DHCPEngine>>::iterator entry(_dhcpInterfaces.find(interfaceName));
 
-                if (entry != _dhcpInterfaces.end()) {                    
-
+                if (entry != _dhcpInterfaces.end()) {
+                    entry->second->StopWatchdog();
                     entry->second->GetIP();
                     result = Core::ERROR_NONE;
                 }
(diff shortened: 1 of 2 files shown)


pwielders:
@mipo57 Could you review this. I am afraid that if we do not signal the failure another issue pops up. So can you please review this.

mipo57:
True, there is an issue just as described. However,as pointed by @pwielders removing RequestFailed(), and most importantly - removing resulting GetIP() would end up in stopping the DHCP process on request failure (a failure that is normal and often expected). 
This situation could happen eg. when device is moved to another network. In such a scenario REQUEST is sent to the old server and after obvious fail, DHCP should find another IP source (eg. request to another server or doing DISCOVER).  If we remove this GetIP() from the process, we will end up without a leased IP. 
I think that it would be a good idea to handle such a situation 100% inside DHCPEngine, without direct NetworkControl involvement (as this is middle of the lease process). 
In any case, after repeated REQUEST timeout, DHCPEngine should end up in a state where REQUEST of another offer or DISCOVER will be made.

HaseenaSainul:
Thanks @mipo57, will make the changes

mipo57:
Looks good now :)

HaseenaSainul:
@mipo57, I have remover StopWatchdog calling sequence from the Dispatcher sequence and its working fine now. Could you please look into it

PR #47, opened 2020-04-20
Added GLIB API support to WebKitBrowser plugin for loading WebKit 2.28.

neeraj-deshpande:


WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -506,7 +525,12 @@ static GSourceFuncs _handlerIntervention =
             , _URL()
             , _dataPath()
             , _view()
+#ifdef WEBKIT_GLIB_API
+            , _guid(g_dbus_generate_guid())
```
pwielders, line 529: We do not want to use any dbus stuiff definitly not gor the generation of 16 random bytes. Suggest to use the high precision clock for example to generate a unique ide. I do not think this gets stored,

pwielders:
@neeraj-deshpande : Can we limit the scoping of the preprocessor defines to the C-functions?
If you look at:
https://github.com/rdkcentral/ThunderNanoServices/blame/wpe-2.28-glib-api/WebKitBrowser/WebKitImplementation.cpp#L724-L756
I think the retrieval of the URL can be done in the new function if you would extract the URL there and just call the already existing functions.
This will prevent the large amount of code duplication you have now.

neeraj-deshpande:
Sure @pwielders   I will do the change.

pwielders:
@neeraj-deshpande what is the status of this pull request ? 
Did you test it with the current WebKit and the 2.28 webkit? Are both compiling and working ?
If so, can we merge it ?

neeraj-deshpande:
@pwielders It also works with current WebKit. I will test it again today.

neeraj-deshpande:
@pwielders Custom JS functions are not yet working and JIRA task is created for it.

pwielders:
> @pwielders Custom JS functions are not yet working and JIRA task is created for it.

Is it not woring with the current setup (CLIB API or is it not working for th new setup (GLIB API).
If it is  the later I still suggest to continue and create a new request for that part.

It is iportant that with this pull request we do not break the current functionality (so the workings of the CLIB API)

pwielders:
Sorry wrong button

neeraj-deshpande:
@pwielders I tested the pull request with WebKit 2.22 and WebKit 2017. It doesn't break.

It requires latest commit of WebKit 2.22 and WebKit 2017 (as Enrique added C API WKPreferencesSetMediaContentTypesRequiringHardwareSupport to WebkitBrowser plugin).

Custom JS functions are not working with GLIB API. WPEU228-5 JIRA task is created for it.

PR #62, opened 2020-05-09
[COMRPC] Example between two applications using COMRPC.

pwielders:


examples/SimpleCOMRPCClient/SimpleCOMRPCClient.cpp
```diff
@@ -23,6 +23,34 @@
...
+    Math& operator= (const Math&) = delete;
+
+    Math() {
+    }
+    ~Math() override {
+    }
+
+public:
+    // Inherited via IMath
+    virtual uint32_t Add(const uint16_t A, const uint16_t B, uint16_t& sum) const override
```
MFransen69, line 38: Better not to use both virtual and override at the same time

examples/SimpleCOMRPCClient/SimpleCOMRPCClient.cpp
```diff
@@ -23,6 +23,34 @@
...
+    {
+        sum = A + B;
+        return (Core::ERROR_NONE);
+    }
+    virtual uint32_t Sub(const uint16_t A, const uint16_t B, uint16_t& sum) const override
```
MFransen69, line 43: see comment above

examples/SimpleCOMRPCInterface/ISimpleCOMRPCInterface.h
```diff
@@ -50,5 +57,15 @@ namespace Exchange {
...
 
+    // This is an example to show the workings and how to develope a COMRPC/JSONRPC method/interface
+    struct IMath : virtual public Core::IUnknown {
+
+        enum { ID = ID_MATH };
+
+        virtual ~IMath() {}
+
+        virtual uint32_t Add(const uint16_t A, const uint16_t B, uint16_t& sum /* @out */)  const = 0;
+        virtual uint32_t Sub(const uint16_t A, const uint16_t B, uint16_t& sum /* @out */)  const = 0;
```
MFransen69, line 68: Matter of taste of course but I would use Substract in full, more clearly indicates the purpose as Sub has different meanings

examples/SimpleCOMRPCServer/SimpleCOMRPCServer.cpp
```diff
@@ -224,11 +225,13 @@ class COMServer : public RPC::Communicator {
...
         const Core::NodeId& source,
-        const string& proxyServerPath)
+        const string& proxyServerPath,
+        const Core::ProxyType< RPC::InvokeServerType<1, 0, 4> >& engine)
         : RPC::Communicator(
             source, 
             proxyServerPath, 
-            Core::proxy_cast<Core::IIPCServer>(Core::ProxyType< RPC::InvokeServerType<1, 0, 4>>::Create()))
+            Core::proxy_cast<Core::IIPCServer>(engine))
+        , _remoteEntry(nullptr)
```
MFransen69, line 234: Totally not important ;) but for consistency with our code base better to have the , on the end of the previous line

examples/SimpleCOMRPCServer/SimpleCOMRPCServer.cpp
```diff
@@ -310,17 +332,20 @@ int main(int argc, char* argv[])
     Core::NodeId comChannel;
     string psPath;
 
+    printf("\npierre is old and builds a COMServer for funs and giggles :-)\n");
+
     if (ParseOptions(argc, argv, comChannel, psPath) == true) {
-        printf("\npierre is old and builds a COMServer for funs and giggles :-)\n");
```
MFransen69, line 314: Haha, never knew this was even in here ;)

woutermeek: I might know who added that 🙈

examples/SimpleCOMRPCClient/SimpleCOMRPCClient.cpp
pwielders: Agree and since this is a show and tell example,let me remove the virtual.

examples/SimpleCOMRPCClient/SimpleCOMRPCClient.cpp
pwielders: See response above :-)

examples/SimpleCOMRPCInterface/ISimpleCOMRPCInterface.h
pwielders: Indeed, matter of tatse casue if it would become Subsract I would use Addition. But since now they are both treated as the mnemonic it has consitent naming, so leave it like this.

examples/SimpleCOMRPCServer/SimpleCOMRPCServer.cpp
pwielders: Style is always debatable. I think it is good for consistency so it was agreed that we will run the WebKit Style. 
The WebKit is initializing it in this way, I think, I am not even sure but I saw it a lot and we are already doing it a lot like this. So I think this is conform standard. 
If it is not the WebKit Style, suggest we run the transform with the WebKit style once more to align this over all code.

examples/SimpleCOMRPCServer/SimpleCOMRPCServer.cpp
pwielders: Now someone knos how long it took this code to gte into the archive :-)

PR #65, opened 2020-05-14
Bluetooth Improvements

sebaszm:
Adds:
- async pairing process,
- remote device class and supported uuids retrieving,
- passkey pairing,
- numerical comparison pairing,
- legacy/pin-code pairing,
- include abortpairing in JSON-RPC,
- connect/pairing state machine bugfixes,
- further auto-connect issues fixed,
- first steps of device lookup by address & type,
- more verbose event handlers,
- general code clean up,
- doc update.


BluetoothControl/BluetoothControl.config
```diff
@@ -4,6 +4,8 @@ map()
    kv(interface 0)
    kv(port "/dev/ttyAMA0")
    kv(baudrate 115200)
+   kv(class "0x000424")
```
pwielders, line 7: This was already a bit cryptic, but the handling even shows more cryptic messages :-) Looks like this should be 2 enums? from the code at least. Can we towards the user not make this two or one enum that means something?

BluetoothControl/BluetoothControl.h
```diff
@@ -48,6 +48,12 @@ class BluetoothControl : public PluginHost::IPlugin
                 : _text(Core::ToString(text))
             {
             }
+            ManagementFlow(const uint16_t opcode, const TCHAR* opcodeStr,  const Bluetooth::Address& address, const uint8_t type)
+            {
+                ASSERT(type < 3);
+                static const TCHAR* typeStr[] = { _T("BREDR"), _T("LE_PUBLIC"), _T("LE_PRIVATE") };
```
pwielders: Maybe good to implement (use) the Core::EnumerateType<> template here and define the lookup table for these values in the bleutooth lib so all conversions later on will be equal.

BluetoothControl/BluetoothControl.h
```diff
@@ -432,69 +568,274 @@ class BluetoothControl : public PluginHost::IPlugin
...
-                        const evt_disconn_complete* info = reinterpret_cast<const evt_disconn_complete*>(data);
-                        TRACE(ControlFlow, (_T("EVT_DISCONN_COMPLETE (status:%d, handle:%d, reason:%d)"), info->status, info->handle, info->reason));
-                        if ((info->status == 0) && (_parent != nullptr)) {
-                            DeviceImpl* entry = _parent->Find(info->handle);
-                            if (entry != nullptr) {
-                                entry->Disconnection(info->reason);
-                            }
-                        }
+                    case EVT_EXTENDED_INQUIRY_RESULT:
+                        Update(reinterpret_cast<const extended_inquiry_info*>(data + 1)); // skip number of inquiries which is always 1
```
pwielders: I do like this setup, but just out of curiosity, why not use a reference than on the signature. Now you are suggetsing that it might be a nullptr, so officially you would have to validate for nullptr in the method?

BluetoothControl/BluetoothControl.h
```diff
@@ -432,69 +568,274 @@ class BluetoothControl : public PluginHost::IPlugin
...
-                        TRACE(ControlFlow, (_T("EVT_IO_CAPABILITY_RESPONSE: %s"), host.ToString().c_str()));
-                        if (_parent != nullptr) {
-                            _parent->Capabilities(host, info->capability, info->authentication, info->oob_data);
-                        }
+                    default:
+                        TRACE(ControlFlow, (_T("UNKNOWN EVENT: %02X"), eventData.evt));
                         break;
+                }
+            }
+            void Update(const extended_inquiry_info* data)
```
pwielders, line 723: Here a refernce whould avoid that officially you have to validate for nullptr's :-)

pwielders: And if you are  thinking about changing this, would it be an Idea to alos include the tracing stuff (by signature) in the Flow ? So enhance the ControlFlow, to also includeConstructors that take the ControlFlow(const extended_inquiry_info&) and genrate a proper log line on this ?

BluetoothControl/BluetoothControl.h
```diff
@@ -432,69 +568,274 @@ class BluetoothControl : public PluginHost::IPlugin
...
+                TRACE(ControlFlow, (_T("  capability=%i, authentication=%i, oob_data=%i"),
+                                    data->capability, data->authentication, data->oob_data));
+
+                UpdateDevice<DeviceImpl>(address, [&](DeviceImpl* device) {
+                    device->Capabilities(data->capability, data->authentication, data->oob_data);
+                });
+            }
+            void Update(const evt_user_confirm_request* data)
+            {
+                Bluetooth::Address address(data->bdaddr);
```
pwielders, line 848: Performance thingu, if you only need the address in the trace, place it in the trace. If the trace is not enabled, theis code is not executed.

BluetoothControl/BluetoothControl.h
```diff
@@ -811,13 +1290,23 @@ class BluetoothControl : public PluginHost::IPlugin
                 , _timeout(0)
                 , _callback(nullptr)
                 , _updateJob(*this)
-                , _autoConnectJob(Core::ProxyType<AutoConnectJob>::Create(this))
                 , _autoConnection(false)
+                , _updateJob(Core::ProxyType<UpdateJob>::Create(this))
```
pwielders, line 1377: Don't know of this would make life easier but the core now has Core::Thread::JobType<> and Core::WorkerPool::JobType<> That might make the coding here easier?

PR #67, opened 2020-05-14
Westeros: Initialize WstGLContext and Get WstGLSetDisplayMode function pointer based on the platform support.

HaseenaSainul:
1. WstGLInit and WstGLTerm usage made undefined error while loading this library. Hence linked westeros_gl.so to resolve that issue
2. WstGLSetDisplayMode is only available for westeros-soc-drm, as well the usage of WstGLInit in other westeros-soc (rpi/brcm) make issue, since its creating new context for rpi/brcm cases. So made the context init and set display mode usage invoke only if these are available for a platform soc


Compositor/lib/Wayland/Westeros.cpp
```diff
@@ -419,46 +419,56 @@ namespace Implementation {
...
         WesterosGL(const WesterosGL&) = delete;
         WesterosGL& operator= (const WesterosGL&) = delete;
 
         WesterosGL() 
             : _context(nullptr)
             , _resolution(Exchange::IComposition::ScreenResolution_1080i50Hz) {
-            _context = WstGLInit();
+            Core::Library library(WesterosGlLibName);
+            if (library.IsLoaded() == true) {
+                _wstGLSetDisplayMode = reinterpret_cast<SetDisplayMode>(library.LoadFunction(_T("_WstGLSetDisplayMode")));
```
pwielders, line 434: Is the name of the function not wstGLSetDisplayMode, i.s.o. _WstGLSetDisplayMode ?

Compositor/lib/Wayland/Westeros.cpp
```diff
@@ -419,46 +419,56 @@ namespace Implementation {
     };
  
     class WesterosGL {
+    private:
+        static constexpr TCHAR* WesterosGlLibName = _T("libwesteros_gl.so.0.0.0");
```
pwielders, line 423: Can we make this configurable, just a bit below, there is already a configuration section (https://github.com/rdkcentral/ThunderNanoServices/blob/f52802634deb25784520150043940ab853ab1fb9/Compositor/lib/Wayland/Westeros.cpp#L391-L419) if you add there the default to be as stated here, we can always change the name in the future if the library is renamed.

PR #68, opened 2020-05-14
Compositor:Wayland:InputSelection support added

HaseenaSainul:
Currently the VirtualInput Switch is not get used by Wayland case, hence the SetInput has to be call specifically for Wayland. That support is added

### DIFF
--- a/Compositor/Compositor.cpp
+++ b/Compositor/Compositor.cpp
@@ -147,6 +147,7 @@ namespace Plugin {
         , _composition(nullptr)
         , _service(nullptr)
         , _connectionId()
+        , _inputSwitch(nullptr)
         , _inputSwitchCallsign()
 
     {
@@ -567,7 +568,6 @@ namespace Plugin {
             result = _inputSwitch->Select(callsign);
         }
         else {
-
             Exchange::IInputSwitch* switcher = _service->QueryInterfaceByCallsign<Exchange::IInputSwitch>(_inputSwitchCallsign);
             if (switcher != nullptr) {
                 result = switcher->Select(callsign);
--- a/Compositor/Compositor.cpp
+++ b/Compositor/Compositor.cpp
@@ -148,7 +148,6 @@ namespace Plugin {
         , _service(nullptr)
         , _connectionId()
         , _inputSwitchCallsign()
-        , _topHasInput(false)
 
     {
         RegisterAll();
@@ -192,8 +191,8 @@ namespace Plugin {
 
             _composition->Configure(_service);
 
+            _inputSwitch = _composition->QueryInterface<Exchange::IInputSwitch>();
             _inputSwitchCallsign = config.InputSwitch.Value();
-            _topHasInput = config.TopHasInput.Value();
         }
 
         // On succes return empty, to indicate there is no error text.
@@ -214,6 +213,10 @@ namespace Plugin {
             subSystems->Release();
         }
 
+        if (_inputSwitch != nullptr) {
+            _inputSwitch->Release();
+            _inputSwitch = nullptr;
+        }
         if (_composition != nullptr) {
             _composition->Release();
             _composition = nullptr;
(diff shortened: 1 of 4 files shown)


Compositor/Compositor.cpp
```diff
@@ -148,7 +148,6 @@ namespace Plugin {
         , _service(nullptr)
         , _connectionId()
         , _inputSwitchCallsign()
-        , _topHasInput(false)
```
pwielders, line 151: Lets initialize _inputSiwtch(nmullptr) as well ;-)

Compositor/lib/Wayland/Wayland.cpp
```diff
@@ -402,6 +402,48 @@ namespace Plugin {
...
+            }
+            return status;
+        }
+        uint32_t Consumer(const string& name, const mode) override
+        {
+            return (Core::ERROR_NONE);;
+        }
+        uint32_t Select(const string& name) override
+        {
+            bool status = Core::ERROR_UNAVAILABLE;
```
pwielders, line 434: Here you need to use the "WstCompositorFocusClientByName(_compositor, name);" to have westeros select where the input will be directed to.

Compositor/lib/Wayland/Wayland.cpp
```diff
@@ -402,6 +402,48 @@ namespace Plugin {
...
+            std::list<Entry*>::const_iterator index(_clients.begin());
+            while (index != _clients.end()) {
+                container.push_back((*index)->Name());
+                index++;
+            }
+            return (Core::Service<RPC::StringIterator>::Create<RPC::IStringIterator>(container));
+        }
+        bool Consumer(const string& name) const override
+        {
+            bool status = false;
```
pwielders, line 417: I suggest always returning true here. In westeros case, there is always 1 pipeline, alwyas open (if you close it no-one, no wayland app) would get input.

Compositor/lib/Wayland/Wayland.cpp
```diff
@@ -402,6 +402,48 @@ namespace Plugin {
...
+                if (name == (*index)->Name()) {
+                    status = true;
+                    break;
+                }
+                index++;
+            }
+            return status;
+        }
+        uint32_t Consumer(const string& name, const mode) override
+        {
```
pwielders, line 429: Suggest returning a Core::ERROR_UNAVAILABLE error here.

Compositor/lib/Wayland/Wayland.cpp
pwielders: Was just digger a bit deeper, I think what you want to call here is a single line called _server->SetInput(Name)

pwielders: Did a bit more digging, what you can do here is just:

uint32_t Select(const string& name) override
{
    return (_server->SetInput(name));
}

Compositor/lib/Wayland/Wayland.cpp
pwielders: All in all, these methods will never be called from the Compositor Plugin

HaseenaSainul: Ok, so I will keep this interface as empty with return true;

Compositor/lib/Wayland/Wayland.cpp
HaseenaSainul: Will remove tje SetInput from client and client comparison code also form the Select. Instead will call _server->SetInput(name) as you suggested

Compositor/lib/Wayland/Wayland.cpp
HaseenaSainul: Will change

Compositor/lib/Wayland/Wayland.cpp
HaseenaSainul: I have not removed it fully, since _server->SetInput(name) : ->WstCompositorFocusClientByName is void function. Hence there is no option to validate requested client is valid or not.

pwielders: Agree that this does a bit of checking, although since this is a "concurrent" system a surface you might find now does not nesseceraly have to be available if you do SetInput since if you found it exists, the CPU might be pre-empted, surface destructed, and after the pre-emption, if it continuous it still does not succeed :-)
But at least it gives a bit of security.
But still for simplicity, I suggest *not* to move to the Entry (Client) and from the Entry(Client) to the _Server->SetInput(passing the client name), but directly from here to the Server->SetInput(name), passing the name you received here.
Than we can remove (cleanup, less lines of code) the SetInput on the Entry.

PR #72, opened 2020-05-19
WebKitBrowser: Add console.log support

lilymeng-brcm:


woutermeek:
@MFransen69  @pwielders I would assume we want this to be a trace rather than always on, or would you like to have it always enabled?

WebKitBrowser/WebKitImplementation.cpp
```diff
@@ -193,6 +194,18 @@ namespace Plugin {
...
+        nullptr, // runJavaScriptConfirm
+        nullptr, // runJavaScriptPrompt
+        nullptr, // checkUserMediaPermissionForOrigin
+        nullptr, // runBeforeUnloadConfirmPanel
+        nullptr, // fullscreenMayReturnToInline
+        // willAddDetailedMessageToConsole
+        [](WKPageRef, WKStringRef source, WKStringRef, uint64_t line, uint64_t, WKStringRef message, WKStringRef, const void* clientInfo) {
+            string src = WKStringToString(source);
+            string msg = WKStringToString(message);
+            fprintf(stderr, "[%s:%llu]: %s\n", src.c_str(), line,  msg.c_str());
```
MFransen69, line 207: Please do not use a printf to log information, we have a trace mechanism for this. Please see here how this could be done: https://github.com/rdkcentral/ThunderNanoServices/blob/0987d8003c9b104a1eb5cc1f3ccfb5e9e7d4a8f0/WebKitBrowser/WebKitImplementation.cpp#L566

pwielders:
This is a broader issue than is usggested here. This is due to the fact that originally we where pushing the console log through the InjectableBundle, but if the Injectable bundle is not available (or loaded) the console logging is not being published. So the issue is correct, but we need to broaden it a bit more.
Remove the Traces from Console log from the injectable bundle here:
https://github.com/rdkcentral/ThunderNanoServices/blob/master/WebKitBrowser/InjectedBundle/main.cpp#L267-L278
Suggest to create a dedicated Trace category, like done here:
https://github.com/rdkcentral/ThunderNanoServices/blob/master/WebKitBrowser/HTML5Notification.h#L27-L61
Call this class BrowserConsoleLog (or something identifying the purpose of this trace)
And than Trace it as done here, to avoid expensive parsing of the Trace is turned off:
https://github.com/rdkcentral/Thunder/blob/c8c91a18313b28392ce01e0a193143c5b47f100b/Source/WPEFramework/PluginServer.cpp#L362-L385
so in pseaudo code it would be like:
[](WKPageRef, WKStringRef source, WKStringRef, uint64_t line, uint64_t column, WKStringRef message, WKStringRef, const void* clientInfo) {
        TRACE(BrowserConsoleLog , (message, line, column);
}

Where the constructor of the BrowserConsoleLog  would be something like:
BrowserConsoleLog ::BrowserConsoleLog (const WKStringRef&, const uint64_t line, const uint64_t column)
        _text = '[' + Core::NumberType<uint64_t>(line).Text() + ',' + Core::NumberType<uint64_t> + ']'
                   +WebKit::Utils::WKStringToString(message);

        const uint16_t maxStringLength = Trace::TRACINGBUFFERSIZE - 1;
        if (_text.length() > maxStringLength) {
            _text= messageString.substr(0, maxStringLength);
        }
}

This will avoid 'two" copies of the same ConsoleLine in all WebKitBrowser login is turned on, and since we created a dedictaed category, called  "BrowserConsoleLog " you could add this to be default on (only these traces) during debugging.

PR #80, opened 2020-05-25
Development/containers plugin api sync

mipo57:
Synchronize with latest API changes


ProcessContainers/ContainersJsonRpc.cpp
```diff
@@ -120,11 +120,11 @@ namespace Plugin {
     uint32_t Containers::get_containers(Core::JSON::ArrayType<Core::JSON::String>& response) const
     {
         auto& administrator = ProcessContainers::IContainerAdministrator::Instance();
-        std::vector<string> containers = administrator.Containers();
+        auto containers = administrator.Containers();
 
-        for (auto container : containers) {
+        while(containers->Next()) {
```
MFransen69, line 125: For consistency and readability please also make it while(containers->Next() == true)

ProcessContainers/ContainersJsonRpc.cpp
```diff
@@ -207,11 +206,11 @@ namespace Plugin {
         if (container != nullptr) {
             auto cpuInfo = container->Cpu();
             
-            response.Total = cpuInfo.total;
+            response.Total = cpuInfo->TotalUsage();
             
-            for (auto core : cpuInfo.cores) {
+            for (int i = 0; cpuInfo->CoreUsage(i) != UINT64_MAX; i++) {
```
MFransen69: We could now use NumberOfCores() here I guess

PR #83, opened 2020-05-26
Encrypted content streamer

krp97:
### Brief
This PR brings in an additional Streamer implementation - CENC. It should be verified and merged [along with this PR](https://github.com/WebPlatformForEmbedded/buildroot/pull/195).

This streamer implementation consists of two main libraries:
- Thunder plugin (CENC) 
- Gstreamer plugin (gstcencdecrypt) - contains a custom transform element, which handles the decryption of encrypted container files.

## CENC
Serves the purpose of instantiating a generic pipeline capable of playing the selected content. It exposes a Streamer API (not fully functional), which can be used for manipulating the stream.
On top of standard debugging messages relayed from the gstreamer pipeline, this plugin also provides means for generating gstreamer dot files.

## gstcencdecrypt 
A gstreamer element, about which [you can find out more here.](https://github.com/WebPlatformForEmbedded/gstcencdecryptor/tree/development/init/1)


mhughesacn:
Hi,
The blackduck failure is happening for a few reasons:
- Code match which is essentially boilerplate - happy with this.
- Search found GST_PLUGIN_DEFINE.  This is using LGPL which is correct here.  However, RDK's approach to GStreamer plugins is that all files which comprise the plugin must be licensed as LGPL, and therefore the headers on these files should be LGPL instead of Apache.

- Warning: Licensing a component with parts licensed different to the main Apache license is under active discussion with Comcast Legal, and has not yet been approved.  It is possible that the gstreamer plugin here may have to go in a separate repository.  Otherwise, if approved, we may request some license file rework (but not code changes) in the near future.

Martin
Martin Hughes RDK CMF Compliance  Team

Appendix: LGPL header is as follows:
/*
* Copyright <year> RDK Management
*
* This library is free software; you can redistribute it and/or
* modify it under the terms of the GNU Library General Public
* License as published by the Free Software Foundation, version 2
* of the license.
*
* This library is distributed in the hope that it will be useful,
* but WITHOUT ANY WARRANTY; without even the implied warranty of
* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
* Library General Public License for more details.
*
* You should have received a copy of the GNU Library General Public
* License along with this library; if not, write to the
* Free Software Foundation, Inc., 51 Franklin St, Fifth Floor,
* Boston, MA 02110-1301, USA.
*/

mhughesacn:
Hi Krystian,
Have you added LGPL headers to all files here which are part of the plugin?  So everything here that is part of the plugin's .so file is now headered as LGPL?

Our tools do not handle this very well at the moment so there will be more complaints.  (It is a new situation for the tools etc.)

Thanks.

krp97:
@mhughesacn Thanks for pointing that out, but there still seems to be a problem with copyright flagging it as an invalid license header? On top of that I don't seem to understand, why blackduck is failing. All files used for building the gstreamer plugin(source files, build scripts), have the LGPL prefix that you've specified.

mhughesacn:
Hi Krystian, Thanks for making the changes.  I approved in blackduck, but our other scan tools don't handle the mixed copyright well, so changes to the LGPL files will complain for now.  (I will try to fix this soon, but it is a new situation which has arisen in this and other components.)  For now, I'll approve manually as soon as I see it.  Just to be aware than Comcast Legal may torpedo this arrangement and require some code reorg.  It is under active discussion. - Martin

krp97:
Hi @mhughesacn, thanks for helping out and I'll keep this in mind. I can definitely separate the gstreamer plugin, shall a need for it arise.

pwielders:
@krp97 (Kristian) Maybe it is good to already make a dedicated lib/so for the gstreamer decryptor element as that was the long term goal anyway (our next step). Lets take that step now. For now we will move this component/part (with different license) to our public WPEPlatformForEmbedded repository (temporary) till ComCast decided on the repository for the LGPL code and than we can move it there.

krp97:
Yeah it's already a shared library since I thought we might want to move it somewhere. Let's move it then and I'll adjust the recipes and build scripts to find it.

PR #85, opened 2020-05-27
[DIALServer] Fixes to align the DIALServer with the DIALSpec (2.2)

pwielders:
For this purpose a new flag has been added, "runtimechange" on AppSection to diffrentiate between apps that can change the URL after being started (WebKit -> Result 200) and apps that con not (Cobalt -> Result 501)


Damian-Danylko:
Looks good to me.

PR #88, opened 2020-05-29
[DeviceIdentification] Amlogic implementation changes

arun-madhavan-013:
Reason for change: Added mfrlib based Amlogic implementation for device identiifcation plugin.
Test Procedure: Build DeviceIdentification plugin.
Risks: Low.

Signed-off-by: Arun P Madhavan <arun_madhavan@comcast.com>


DeviceIdentification/Implementation/Amlogic/Amlogic.cpp
```diff
@@ -0,0 +1,103 @@
...
+    {
+        return _chipset;
+    }
+    const string FirmwareVersion() const override
+    {
+        return _firmwareVersion;
+    }
+
+    // Identifier interface
+    uint8_t Identifier(const uint8_t length, uint8_t buffer[]) const override
```
pwielders, line 60: If you supply this interface IIdentifier, make sure you set a proper ID. This ID is used by multiple other components in the system to identify the STB. If you do not implement it, the ID will be the MAC address of the first NIC it finds during netlink (kernel query) but do not provide an implementation that return a 0.

arun-madhavan-013: Thanks; will modify as suggested. Can it be device serial number ?

pwielders: Yes, as long as it something that identifies the box :-)

arun-madhavan-013: Even if i add a valid ascii data say "ABCD" in **buffer** of Identifier(const uint8_t, uint8_t **buffer**[]), I'm expecting to get "ABCD" properly in JSONRPC response but I don't. Seems like its picking some jumbled characters.
Is it expected ?

DeviceIdentification/Implementation/Amlogic/Amlogic.cpp
```diff
@@ -0,0 +1,127 @@
...
+    {
+        uint8_t result = 0;
+        if ((_length != 0) && (_status == mfrERR_NONE)) {
+            result = (_length > length ? length : _length);
+            ::memcpy(buffer, reinterpret_cast<const uint8_t*>(&_identity[0]), result);
```
arun-madhavan-013, line 66: mfrlib is populating **buffer** with **"280b4000011d2800"** but in the JSONRPC response **deviceid** is **not** **"280b4000011d2800"**; it something else.

> [Amlogic.cpp:66]: Notification: buffer = **280b4000011d2800**            <--------- Removed respective code since its from c'tor.
> [SystemInfo.h:329]: Startup: EVENT: Identifier: OEMjgwYjQwMDAwMTFkMjgwMA
> 
> {"jsonrpc":"2.0","id":1234567890,"result":{"firmwareversion":"2.0","chipset":"Amlogic S905X2",**"deviceid":"OEMjgwYjQwMDAwMTFkMjgwMA"**}}

pwielders: The interface allows for binary values (we do not want to enforce any requiremtn on the SystemIdentifier) to be retuned.
So the value you return her is expected to be a binary cvalue. Given your exampl, I think, the device identifier is a binary value.
[28][0b][40][00][01][1d][28][00] (so I think your ID is actually 8 bytes long, probably a UInt64_t :-) )
Since JSONRPC is expected to be text strings, it will not expect binary values like you are returning in the interface here, so the system handles this on your behalf to encode it in base64. Always safe to send as a string.
This means (if you want to turn it back into the original source, that you have to base64 it back to its original value.
If we try to decode you string back it looks like it is failing, but I think that is due to your construction.
The reverse encoding of the Base64 leads to a different value but that is probably due to the way you copy the string. You are, I think, currently copying the stack varaible address with remainder stuff behind it...
So also suggets you do the memcpy properly by stating:
::memcpy(buffer, _identity.c_str(), result);
I am also not sure if the operation on line 109,110. I would suggest (pottentiall the contens could be binary) storing it like:
_identity = string(mfrSerializedData.buf,_identity.length());

This will make sure that if it is binary and the first character = '\0' still all chracters will be stored in the string.

pwielders: While you are changing, would like to know what the length is of the MFR Identifier, is it a Uint64_t? Could it be that the MFRApi is already turning it into a Hexacdecimal string. If you look at the identifier trhat you show it definitiely looks like a Binary number (0..9 and a..f)

pwielders: Ohh by the way, the first three characters (OEM) we (Metrological) reserve to identifiy the kind of vender. It can be set during compilation.

arun-madhavan-013: mfrAPI fetching serial number(107) returns a NULL terminated c-string which here can be considered equivalent to `char data[] = "280b4000011d2800";`
This can be copied to std::string variable with Operator = without any problem(at 109/110) if I'm not mistaken making it text string.

This was the code snippet used to print the attached debug log. Hope this makes clear how the data is encoded.
```
uint8_t Identifier(const uint8_t length, uint8_t buffer[]) const override
{
        uint8_t result = 0;
        char tempbuffer[64] = {'\0'};
        if ((_length != 0) && (_status == mfrERR_NONE)) {
            result = (_length > length ? length : _length);
            ::memcpy(buffer, reinterpret_cast<const uint8_t*>(&_identity[0]), result);
            ::memcpy(data, reinterpret_cast<const uint8_t*>(&_identity[0]), result);
            SYSLOG(Logging::Notification, (_T("buffer = %s"), tempbuffer));
        }
        ..................
}
```
I'm not sure why its again being encoded.

I found that if "280b4000011d2800" when base64encoded results in "MjgwYjQwMDAwMTFkMjgwMA==" which is decodable with any online tool.

If we compare this with JSONRPC result `"deviceid":"OEMjgwYjQwMDAwMTFkMjgwMA"` the actual base64encoded result is prepended with "OE" and trailing "==" is removed in Thunder logic making the result not decodable.

pwielders: If we compare this with JSONRPC result "deviceid":"OEMjgwYjQwMDAwMTFkMjgwMA" the actual base64encoded result is prepended with "OE" and trailing "==" is removed in Thunder logic making the result not decodable.

The trailing '==' is called padding and is **not mandatory**, it is optional. If indeed the current system is set to OE that is incorrect. We always expect this to be 3 characters (so should be OEM, requires to be changed). Decoding on https://cryptii.com/pipes/base64-to-hex, using MjgwYjQwMDAwMTFkMjgwMA (no padding needed as expected) results in 280b4000011d2800 (text)Which is indeed what we put in.

Circling back to the construction:
 ::memcpy(buffer, reinterpret_cast<const uint8_t*>(&_identity[0]), result);
it is still a very dangerous construction. It might work if, and only if,  the implementation of the std string class returns a reference to the physical character in the string. The specification does not mandate it, so if the implementation wants to copy this to a static location and return a reference to only the single character, it is still according to specification and it is valid from the calling perspective, but the ::memcpy will copy the incorrcte stuff.
That is why I suggest you use the functions suitable for this purpose:
http://www.cplusplus.com/reference/string/string/c_str/
wich returns, as the description states an array of charcaters. It also avoids using a cast. Which is always good.
This will make the code portable (agnostic to the impleemntation of std::string::operator[]) and maintainable (ohh they are expecting an array and not 1 character)!

I did not examine the MFR API with respect to the extraction of the Identifier but If, as you state, you know it is a readable (ansi/UTF8) string coming from the MFR library, than there is no need to store the _length field or to do any operation on the length field. And I would suggest removing it. You can always use the std::string::length() method. So to avoid confusion also remove the _lengty member in this whole setup

arun-madhavan-013: Informative...
Updated the implementation to use std::string and removed the unused _length variable.

PR #91, opened 2020-06-02
[NetworkControl] Adapt to netlink changes

mipo57:
Adaptation to changes in core Netlink code. On top of that, removed
callbacks being called in ResourceMonitor Thread. Now they are
submitted to the worker pool. Also changed GetIP to more descriptive AcquireIP


mipo57:
This is mostly a way to sync with Netlink changes in Core (https://github.com/rdkcentral/Thunder/pull/101) and to have it in PR as @MFransen69 requested. This is quite temporary and more changes are needed to fully implement Netlink: 
- Make it fully asynchronous from the plugin side
- Fix problems with the gateway (it's not set correctly, but for some reason assigning 0.0.0.0 to broadcast makes it work...)
- Separate Broadcast & Interface Ip addresses.
- Differentiate between network event types
- Implement DHCP lease extension
- Now every interface is a static one. That makes it invalid as soon as DHCP (or another internal or external call) changes anything. As we get update on IP/Link changes from Netlink, it would be a good idea to use them and have all the info as fresh as possible.

Also, I think it would be a good idea to make more separation between DHCP engine and NetworkController, as now there is a lot of integration between them.

NetworkControl/DHCPClientImplementation.h
```diff
@@ -901,7 +914,19 @@ namespace Plugin {
                             if (xid == _discoverXID) {
                                 AddUnleasedOffer(Offer(source, frame, options));
                                 TRACE(Trace::Information, ("Received an Offer from: %s", source.HostAddress().c_str()));
-                                _discoverCallback(_unleasedOffers.back());
+
```
pwielders, line 917: What you are calling, is it non-deterministic or very lengthy in duration that you would like to decouple it. If none of those are applicable, I would not decouple it as decoupling it is relatively lengthy

mipo57: At the moment, down the line series of Netlink exchange() calls are made - and waittime can't be zero with the current implementation, as they are all dependent on each other. That is very likely to change, as ideal they could be called completely asynchronously as chain of events without any wait-time. However, at time of this sync PR all of these calls are blocking ResourceMonitor

pwielders: oke

PR #93, opened 2020-06-02
Cobalt: suspend set only auto start is true

HaseenaSainul:
Currently the Cobalt is starting always in suspended mode. This creating issue for DIALServer passive case. The DIAL is trying to activate Cobalt, but could not able to manage the app, since its in suspended state. So as like Netflix, better to keep the Cobalt in suspended mode only if it is starting with autostart as true.


HaseenaSainul:
Closing this PR, since we have to give complete control to the JS apps

PR #100, opened 2020-06-05
[IExternal] Aligning to a slimmer interface. Remove and Rename.

pwielders:
Also implemented the publishing/exposure of the IExternals available here to the ICatalog interface.

### DIFF
--- a/IOConnector/IOConnector.cpp
+++ b/IOConnector/IOConnector.cpp
@@ -77,10 +77,12 @@ namespace Plugin
     };
 
     IOConnector::IOConnector()
-        : _service(nullptr)
+        : _adminLock()
+        , _service(nullptr)
         , _sink(this)
         , _pins()
         , _skipURL(0)
+        , _notifications()
     {
         RegisterAll();
     }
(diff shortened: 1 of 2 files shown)


IOConnector/IOConnector.cpp
```diff
@@ -209,43 +211,56 @@ namespace Plugin
...
-    /* virtual */ Exchange::IExternal* IOConnector::Resource(const uint32_t id)
+    /* virtual */ void IOConnector::Unregister(ICatalog::INotification* sink)
     {
+        _adminLock.Lock();
 
-        Exchange::IExternal* result = nullptr;
-
-        // Lets find the pin and trigger if posisble...
-        Pins::iterator index = _pins.find(id);
+        // Make sure a sink is not registered multiple times.
```
MFransen69, line 236: Copy and paste I guess. Although most people will understand that the comment does not reflect the code perhaps better to change this to prevent any confusion

IOConnector/IOConnector.cpp
```diff
@@ -209,43 +211,56 @@ namespace Plugin
...
+        NotificationList::iterator index = std::find(_notifications.begin(), _notifications.end(), sink);
 
-        if (index != _pins.end()) {
-            result = index->second.Pin();
-            result->AddRef();
+        if (index != _notifications.end()) {
+            (*index)->Release();
```
MFransen69, line 240: Although it is in the lock and no chance that index is being used from the notifications list before the next line I would change the order and first remove and then release.

IOConnector/IOConnector.h
```diff
@@ -125,7 +126,8 @@ namespace Plugin {
             std::list<IHandler*> _handlers;
         };
 
-        typedef std::map<const uint32_t, PinHandler> Pins;
+        using NotificationList = std::list< Exchange::IExternal::ICatalog::INotification* >;
+        using Pins = std::map<const uint32_t, PinHandler>;
```
MFransen69, line 130: use unordered_map perhaps or do you really need them ordered?

PR #107, opened 2020-06-08
Modified RemoteControl plugin - delimiter used for looking key code within the designator

neeraj-deshpande:


pwielders:
The @ sign is used to indicate indexes and thus has the special meaning of indicating that it is not part of the designator. So it should be the @ sign from a design/architecture perspective.
I understand that the documentation was not inline with the code, so please adopt the documentation i.s.o. the code :-)

PR #117, opened 2020-06-15
Added initial default values for memorylimit in Monitor plugin

brainstem123:


Monitor/CMakeLists.txt
```diff
@@ -18,6 +18,14 @@
 set(PLUGIN_NAME Monitor)
 set(MODULE_NAME ${NAMESPACE}${PLUGIN_NAME})
 
+set(PLUGIN_WEBKITBROWSER_MEMORYLIMIT "614400")
```
MFransen69, line 21: As just discussed let's make it a cached variable

PR #119, opened 2020-06-16
[NetworkControl] Prevented requesting IP for empty MAC

mipo57:
DHCP issued a request with the MAC assigned at the interface launch. On some boxes, MAC is set to zeroes and updated after the interface was set up.  This resulted in eth0 ending without an IP or long DHCP handshakes (as a fully failed handshake was fist needed to deny the request)


NetworkControl/DHCPClientImplementation.h
```diff
@@ -591,6 +591,11 @@ namespace Plugin {
         {
             return (_interfaceName);
         }
+        inline void UpdateMAC(uint8_t* buffer, uint8_t size) {
```
pwielders, line 594: I think the signature should read like:
inline void UpdateMAC(const uint8_t buffer[], const uint8_t size);

NetworkControl/DHCPClientImplementation.h
```diff
@@ -591,6 +591,11 @@ namespace Plugin {
...
+            ASSERT(size == 6);
```
pwielders, line 595: Suggest removing the magical 6, by ASSERT(size == sizeof(_MAC));

NetworkControl/NetworkControl.h
```diff
@@ -357,12 +357,22 @@ namespace Plugin {
...
-
-                auto offerIterator = _client.UnleasedOffers();
-                if (offerIterator.Next() == true) {
-                    Request(offerIterator.Current());
-                } else {
-                    Discover(preferred);
+                Core::AdapterIterator adapter(_client.Interface());
+
+                if (adapter.IsValid()) {
+                    // Some developer devices assign MAC via script after interface
```
pwielders, line 363: I think the rootcause here is the fact that the NIC only assigns a MAC address, only after the NIC has been activated. I assume that as part of the of the NetworkControl the NIC gets activated. I think it is good to "reload" the MAC address at the moment the NIC is activated.
So my preference would be not to always iterate over the adapters but do this as you "enable" the NOC
For example here:
https://github.com/rdkcentral/ThunderNanoServices/blob/89d34b30dd75eb503383606bd06f9f239f21a2a9/NetworkControl/NetworkControl.cpp#L878
This is where we get the notification that the NIC is up and thus it means that the MAC must be set now. Maybe it is good to only do this exercuise if the current MAC == 0.

mipo57: Just to be clear:
After recent changes, we get the new MAC the moment it is set (by the means of netlink event), and then propagate Event (via Observer) to NetworkControl. As GetIP is only called from reload, it actually should account to exactly the same thing :). 
But I totally agree with remark - it might be unnatural and confusing to call it from GetIP function.

I think it would be a good idea to also pass info what happened to the event call, but that's aside.

PR #124, opened 2020-06-22
WebServer: Decouple the proxy from the plugin's availability

johansaji:
In the original implementation, the proxy for DIALServer and Device info
is enabled by default when these plugins are enabled. There are cases
where we need proxy to be enabled to any one of the component
irrespective of the other plugins state. For example, in a system where
DIAL and DeviceInfo is enabled, and dial is configured to use the proxy server,
Webserver need to expose the port to the external world. In this case,
DeviceInfo is also get exposed to the external world. In order to avoid
this, the config has been tweaked to use a different config variable to
enable the proxy.


WebServer/WebServer.config
```diff
@@ -1,3 +1,6 @@
+option(PLUGIN_WEBSERVER_PROXY_DEVICEINFO "Enable proxy for DeviceInfo" OFF)
```
pwielders, line 1: For backwards compatibility, can we make the default value, the plugin value, so something like:
option(PLUGIN_WEBSERVER_PROXY_DEVICEINFO "Enable proxy for DeviceInfo" ${PLUGIN_DEVICEINFO})

johansaji: Okey

PR #127, opened 2020-06-22
[Networking] WebServer dependency changes

mipo57:
This should fix Splashscreen sometimes failing due to race with network initialization.

Changes:
1. WebServer now depends on Network to be available and UX depends on WebServer.
2. Network precondition is now set whenever all defined interfaces are UP. Previously it 
   was set when the first IP was assigned to any of the interfaces. That could lead to
   ambiguous behaviour and having IP is not critical indicator if network is ready
   (eg. when depending on wifi that has not yet connected to any network)

### DIFF
--- a/NetworkControl/NetworkControl.cpp
+++ b/NetworkControl/NetworkControl.cpp
@@ -217,6 +217,18 @@ namespace Plugin
         // From now on we observer the states of the give interfaces.
         _observer->Open();
 
+        PluginHost::ISubSystem* subSystem = _service->SubSystems();
+
+        ASSERT(subSystem != nullptr);
+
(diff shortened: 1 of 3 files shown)


mipo57:
@pwielders What do you think of the second change? It's quite an important one from a design perspective, as it changes what does it mean to have "network" ready. As an alternative, I think of adding a new kind of precondition like "NetworkInterface". 
However, I guess that having "Network" meaning network interfaces are up & ready and "Internet" meaning there is internet connection is quite a clear definition. Contrary to previous "Network" meaning that some devices got some IP (problematic e.g. in the case described in PR) .

PR #128, opened 2020-06-23
WebServer: Makes file server configurable

johansaji:
This commit is for making the file serving feature configurable.
This is needed since, in cases we only wants the WebServer to be a
proxy for another plugin, we don't want to exposes the filesystem
contents to external world. In such case, we could use this to disable
the file serving feature.
With this change, we would be able to configure WebServer only as a
proxy server.


MFransen69:
To finish this pull request a follow up pull request was created:  https://github.com/rdkcentral/ThunderNanoServices/pull/338

PR #129, opened 2020-06-23
[Power] HP40A-126 : Power plugin - persistent state feature

arun-madhavan-013:
Reason for change: Linux Implementation - Adding new feature of persistent power state (boot up from power loss in standby if device was in standby state).
Require: https://github.com/WebPlatformForEmbedded/meta-wpe/pull/469
Signed-off-by: Arun Madhavan <Arun_Madhavan@comcast.com>

### DIFF
--- a/Power/CMakeLists.txt
+++ b/Power/CMakeLists.txt
@@ -54,7 +54,6 @@ target_include_directories( ${MODULE_NAME}
 
 find_package(BcmPowerManager QUIET)
 find_package(MFRPwrMgrLibs QUIET)
-find_package(MFRFWLibs QUIET)
 
 if(${PLUGIN_POWER_MFRPERSIST_STATE} STREQUAL ON)
     message("Enabling Persistent power state feature...")
--- a/Power/CMakeLists.txt
+++ b/Power/CMakeLists.txt
@@ -55,12 +55,14 @@ target_include_directories( ${MODULE_NAME}
 find_package(BcmPowerManager QUIET)
 find_package(MFRPwrMgrLibs QUIET)
 
-if(${PLUGIN_POWER_MFRPERSIST_STATE} STREQUAL ON)
+if(${PLUGIN_POWER_MFRPERSIST_STATE} STREQUAL ON AND MFRPWRMGRLIBS_FOUND)
     message("Enabling Persistent power state feature...")
-    add_definitions(-DPOWER_PERSIST_STATE=1)
-    if(MFRPWRMGRLIBS_FOUND)
-        target_link_libraries(${MODULE_NAME} PRIVATE pwrmgrlibs::pwrmgrlibs)
-    endif()
+    target_sources(${MODULE_NAME} PRIVATE
+        PowerImplementation/${PLUGIN_POWER_IMPLEMENTATION}/MFRPersistPower.cpp)
+    target_link_libraries(${MODULE_NAME} PRIVATE pwrmgrlibs::pwrmgrlibs)
+else()
+    target_sources(${MODULE_NAME} PRIVATE
+        PowerImplementation/Stub/MFRPersistPowerStub.cpp)
 endif()
 
 if(BCM_PM_FOUND)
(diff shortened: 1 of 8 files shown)


arun-madhavan-013:
> As discussed I think this is a good generic feature and thus should be treated as such. What it means is that the feature should be triggered from the PLugin, not from the implementation. This means that you need to extend the C-Interface, found her:
> https://github.com/rdkcentral/ThunderNanoServices/blob/master/Power/Implementation.h
> with something like:
> 
> EXTERNAL enum WPEFramework::Exchange::IPower::PCState power_get_persisted_state();
> and
> EXTERNAL void power_set_persisted_state(const enum WPEFramework::Exchange::IPower::PCState);
> 
> Than I think you have to change the "EXTERNAL void power_initialize(power_state_change callback, void* userData, const char* config);" with an additional parameter, of type const enum WPEFramework::Exchange::IPower::PCState, because this is the persisted state, retrieved using the new power_get_persisted_state function. This is executed from the plugin initialization, here: https://github.com/rdkcentral/ThunderNanoServices/blob/master/Power/Power.cpp#L67 since this is a new generic feature, and now it becomes available to "all" implementations.
> Than since this is MFR functionality, create a cpp file called MFRPersistPower.cpp in https://github.com/rdkcentral/ThunderNanoServices/tree/master/Power/PowerImplementation/Linux and implement the two new functions form the C interface.
> 
> EXTERNAL enum WPEFramework::Exchange::IPower::PCState power_get_persisted_state() {
> ...
> pmReadPowerState(&lastDevicePowerState);
> ...
> }
> and
> EXTERNAL void power_set_persisted_state(const enum WPEFramework::Exchange::IPower::PCState) {
> ...
> pmSavePowerState
> ...
> }
> 
> for the implemntations where you have no clue how this should look like, implement/add the two nwe function as follows:
> 
> EXTERNAL enum WPEFramework::Exchange::IPower::PCState power_get_persisted_state() {
> // We did not store, as we do not need this feature,
> return (Exchange::IPower::PCState::On);
> }
> and
> EXTERNAL void power_set_persisted_state(const enum WPEFramework::Exchange::IPower::PCState) {
> // No clue where to store or what to do, open for the specific deployment
> }
> 
> There where you now have the preprocessor stuff, just remove it all and iuse the new genric functions. and we are ready for the future and this has become a generic feature (also applicable to the broacom implementation (remeber to add, for now the implementation shown above ther)

We are not sure whether this requirement is a generic one. That is the reason the changes are confined within Implementation section.
Thanks, will try to change as suggested.

pwielders:
> 
> We are not sure whether this requirement is a generic one. That is the reason the changes are confined within Implementation section.
> Thanks, will try to change as suggested.

I am sure :-) The use case presented is a use case that is applicable to any environment/operator. A power failure is always applicable wether we have an AMLogic box or an Broadcom box and if the power comes back up we need to make sure that the connected device (TV) should not start playing. So to check at startup what the state was where the box was left in during the last power state change (persist over boot) is a real valid usecase and would like to make that being triggered from the plugin and not from the implementation.

arun-madhavan-013:
> > We are not sure whether this requirement is a generic one. That is the reason the changes are confined within Implementation section.
> > Thanks, will try to change as suggested.
> 
> I am sure :-) The use case presented is a use case that is applicable to any environment/operator. A power failure is always applicable wether we have an AMLogic box or an Broadcom box and if the power comes back up we need to make sure that the connected device (TV) should not start playing. So to check at startup what the state was where the box was left in during the last power state change (persist over boot) is a real valid usecase and would like to make that being triggered from the plugin and not from the implementation.

:-)

Also; while testing the implementation, we have noticed that Compositor initialization also happens before other plugins and broadcasts **PLATFORM precondition**. Then after initializing the compositor client section it broadcasts **GRAPHICS precondition**. This causes DRM/KMS trigger resulting in HDMI out.
Seems we require some mechanism to detect the Power state and configure DRM/KMS to mute the output.

arun-madhavan-013:
Added a common Stub and added Broadcomm section changes too so that build will not break. ([here](https://github.com/rdkcentral/ThunderNanoServices/pull/129/files#diff-854fbe7d7ea1fc861ee25dbc2d138144)).

PR #130, opened 2020-06-25
[PlayerInfo] Implemented Dolby output switch

krp97:
The PlayerInfo plugin has been extended in order to provide
means for switching dolby output types.


PlayerInfo/PlayerInfo.cpp
```diff
@@ -52,6 +52,17 @@ namespace Plugin {
 
                     _player->Release();
                     _player = nullptr;
+                } else {
+
+                    // The code execution should proceed regardless of the _dolbyOut
+                    // value, as it is not a essential.
+                    // The relevant JSONRPC endpoints will return ERROR_UNAVAILABLE,
+                    // if it hasn't been initialized.
+                    _dolbyOut = service->Root<Exchange::Dolby::IOutput>(_connectionId, 2000, _T("DolbyOutputImplementation"));
```
pwielders, line 61: You probably did not test this out-of-process :-) Which is not needed as this should not run out-of-process, but it would fail ;-)
You are replacing a _connectionId here with a new _connectionId. I think that is not you intention. My suggestion is to use a QI here to get the _dolbyOut interface. The fact it support Dolby or not  can be"retrieved" using the Interface navigation methods of the Player Object you created. This way the whole PlayerInfo object lives in the same process space and thus the _connectionId clash is nolonger an issue and it keeps it all toghether.
Lets discuss thiss offline.

PlayerInfo/Dolby/DolbyOutput.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+                ASSERT(errorCode == Core::ERROR_NONE);
+            };
+
+            BEGIN_INTERFACE_MAP(DolbyOutputImplementation)
+                INTERFACE_ENTRY(Exchange::Dolby::IOutput)
+            END_INTERFACE_MAP
+        };
+
+        SERVICE_REGISTRATION(DolbyOutputImplementation, 1, 0);
```
pwielders, line 51: I saw the way you instantiated this. I think we need to modify this a bit. Lets have a call

PR #131, opened 2020-06-29
FirmwareControl: Change to update download progress to the initiated process

HaseenaSainul:
1. Changes to update download progress
2. Moved Hash value comparison logic to the FirmwareControl context, instead handling it from the webtransfer context (it helps to avoid delay in the json event notification during the Hash comparison)

### DIFF
--- a/FirmwareControl/DownloadEngine.h
+++ b/FirmwareControl/DownloadEngine.h
@@ -34,46 +34,6 @@ namespace PluginHost {
     class DownloadEngine : public Web::ClientTransferType<Core::SocketStream, Web::SignedFileBodyType<Crypto::SHA256>> {
 
     private:
-        class ProgressHandler {
-        public:
-            ProgressHandler()
-                : _parent(nullptr)
-            {
-            }
-            ProgressHandler(const ProgressHandler& progressHandler)
-                : _parent(progressHandler._parent)
-            {
-            }
-            ProgressHandler(DownloadEngine* callback)
-                : _parent(callback)
-            {
-            }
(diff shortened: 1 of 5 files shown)


FirmwareControl/DownloadEngine.h
```diff
@@ -154,6 +218,8 @@ namespace PluginHost {
         Core::CriticalSection _adminLock;
         INotifier* _notifier;
         Core::File _storage;
+        uint16_t _fileSize;
+        Core::TimerType<ProgressHandler> _progressHandler;
```
pwielders, line 259: This, under the hood, construct a thread. I thin we would like to avoid these kind of threads as there is no heavy lifting on this I suggets working with a Workerpool::JobType<ProgressHandler> that you just schedule on X seconds, See WorkerPool::Schedule().

FirmwareControl/DownloadEngine.h
```diff
@@ -79,13 +110,46 @@ namespace PluginHost {
             return (result);
         }
 
+        void StartProgressNotifier(const uint16_t interval)
+        {
+            ProgressHandler entry(*this);
```
pwielders: Using a WorkerPool::JobType<ProgressHandler> makes your life so much easier :-)

FirmwareControl/DownloadEngine.h
HaseenaSainul: Thanks @pwielders for the comments, changed the TimerType to JobType

FirmwareControl/DownloadEngine.h
HaseenaSainul: Done :)

PR #133, opened 2020-06-30
Added keyholdtime & keyrepeattime configuration parameters.

neeraj-deshpande:


Cobalt/CobaltImplementation.cpp
```diff
@@ -39,11 +39,20 @@ class CobaltImplementation:
...
+            , Width(1280)
+            , Height(720)
+            , KeyHoldTime()
+            , KeyRepeatTime()
+            , ClientIdentifier()
+        {
             Add(_T("url"), &Url);
             Add(_T("width"), &Width);
             Add(_T("height"), &Height);
+            Add(_T("keyholdtime"), &KeyHoldTime);
```
pwielders, line 54: Can we use the same JSON names as we use in the RemoteControl, so that people are aware that they are changing the same type of parmaters for this appEngine:
https://github.com/rdkcentral/ThunderNanoServices/blob/master/RemoteControl/RemoteControl.h#L146-L147

PR #138, opened 2020-07-02
TraceControl: update config parameters using build options

HaseenaSainul:


pwielders:
I think we have a backwards compatibility broken. Now the Console and Ssylog will be set, so the TraceControl will not determine where to output as the Configurations is depicting it. I think that is yunwanted behaviour. Now if we runThunder in the forground, the logs will still go to SYSLOG i.s.o. Console. I think that was a usefull feature..

pwielders:
Don't know the exect ticket associated with it, leave it up to MFransen

PR #146, opened 2020-07-07
[WifiControl] Support for specifying WPA protocol type

mipo57:
WPA can now be specyfied to support WPA, WPA2 or both protocols
Also fixed WPASupplicant::Config::CopyProperties to do actual copying


mipo57:
Sidenote: From frontend side https://github.com/rdkcentral/RDKSplashScreen/blob/38d312ebd40b847c6320cbc4d2d8de6824f5d653/dist/web/js/src.es5/appBundle.js#L1140 should be changed to either
```js
if (n.pairs[0].method === 'WPA2' || n.pairs[0].method === 'WPA') type = 'WPA_WPA2';else [...]
```
or
```js
if (n.pairs[0].method === 'WPA2') type = 'WPA2'; else if(n.pairs[0].method === 'WPA') type = 'WPA';else [...]
```

WifiControl/Network.h
```diff
@@ -365,6 +371,8 @@ namespace WPASupplicant {
         bool Hash(const string& hash);
         bool PresharedKey(const string& presharedKey);
         bool Enterprise(const string& identity, const string& password);
+        bool Protocols(const uint8_t protocolFlags); //
```
HaseenaSainul, line 374: minor remark, please remove //

WifiControl/WifiControl.cpp
```diff
@@ -197,6 +197,44 @@ namespace Plugin
         return result;
     }
 
+    uint8_t WifiControl::GetWPAProtocolFlags(const JsonData::WifiControl::ConfigInfo& settings, const bool safeFallback) {
+        uint8_t protocolFlags = 0;
+
+        switch (settings.Type.Value()) {
```
HaseenaSainul, line 203: Just wanted to check, what happen if we try to set WifiControl::TypeType as invalid value through jsonrpc call. I suspect, this will set protocolFlags as 0 and will not report error in this case.

mipo57: That's what safeFallback is for. If it's set to true, it will assign a supplicant's default value of WPA | WPA2. In both cases a warning will be issued

mipo57: Updated warning message to be more expressive

HaseenaSainul: ok @mipo57

PR #154, opened 2020-07-12
[WINDOWS] The windows compiler does not allow dynamic storage on stack.

pwielders:
I thought windows (VS2019) is a pretty modern compiler but the dynamic allocation of arrays on a stack is not allowed. A I missing something. Used the old fashion ALLOCA() but please feedback if we could revert and have it compiling under windows as well..


sebaszm:
True that VLA is non standard C++, but neither is alloca. Though agree on alloca if fits all compilers. (Perhaps here an alternative all-proof solution would just be TCHAR encodedDataUrl[128] coupled with and assert.)

Also see https://godbolt.org/z/5Waj99 for some amusement.

PR #159, opened 2020-07-17
WifiControl: update to avoid deadlock situation and doc/jponrpc interface updates

HaseenaSainul:
1. Currently same lock is used inside the Controller class for connect/disconnect/scan as well SendData/ReceiveData. This will sendup deadlock some times since the calls are happening in the two way direction. Hence removed the locks inside the connect/disconnect/scan of controller class. Also used lock from JsonRPC/ReST calls during the invoke of _controller->connect/disconnect/scan  to ensure there is no parallel calls and end up with some overwrite on the submitted requests.

2. Updated document and WifiControlJsonRPC comments to align with latest WifiControl.json

### DIFF
--- a/WifiControl/Controller.cpp
+++ b/WifiControl/Controller.cpp
@@ -177,166 +177,179 @@ namespace WPASupplicant {
     /* virtual */ uint16_t Controller::SendData(uint8_t* dataFrame, const uint16_t maxSendSize)
     {
         uint16_t result = 0;
-        _adminLock.Lock();
-        if ((_requests.size() > 0) && (_requests.front()->Message().empty() == false)) {
-            string& data = _requests.front()->Message();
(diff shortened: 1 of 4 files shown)


HaseenaSainul:
Thanks @pwielders for the comments. I think adding a new critical section to protect read/write will not avoid the deadlock case. Since the ReceiveData and SendData is already under a critical section from Core:SocketPort, so adding a new one will not give any additional gain. In such case also, the they need to acquire adminLock to access _requests list. The _requests list is already used by the Controller thread to submit each request with adminLock. So it again endup with deadlock situation

HaseenaSainul:
@pwielders, introduced new lock to handle send/receive data and moved the response handling logic to worker pool

HaseenaSainul:
Closing, since it already covered as part of another PR 169

PR #160, opened 2020-07-20
[NetworkControl] DHCP client now properly set 0.0.0.0 as source

mipo57:
According to rfc2131 every DHCP packet before ACK should have 0.0.0.0
set as source in IP frame. It seems that some routers are very strict
about it. However to do that reliably, DHCP socket had to be
reimplemented as raw AF_POCKET type


NetworkControl/DHCPClientImplementation.cpp
```diff
@@ -18,27 +18,169 @@
...
+
+        const bool IsIncomingMessage()
+        {
+            const iphdr* ipHeader = reinterpret_cast<iphdr*>(_buffer);
+            const udphdr* udpHeader = reinterpret_cast<udphdr*>(_buffer + udpHeaderOffset);
+
+
+            return (_bufferSize > minimumFrameSize) 
+                    && (ipHeader->protocol == PROTOCOL_UDP)
+                    && (udpHeader->dest == htons(68));
```
pwielders, line 112: 68 (DHCP port, I guess:-) ) seems to be a magical, place make it a const or a parameter because I think it would be a good idea that the RawPackage is something we could place in core as common functionality ?

NetworkControl/DHCPClientImplementation.cpp
```diff
@@ -18,27 +18,167 @@
...
+        void RecalcHeadersSizes() 
+        {
+            iphdr* ipHeader = reinterpret_cast<iphdr*>(_buffer);
+            udphdr* udpHeader = reinterpret_cast<udphdr*>(_buffer + udpHeaderOffset);
+
+            ipHeader->tot_len = htons(sizeof(iphdr) + sizeof(udphdr) + _dhcpFrameSize);
+            udpHeader->len = htons(sizeof(udphdr) + _dhcpFrameSize);
+        }
+
+        void RecalcIpHeaderChecksum()
```
bramoosterhuis, line 145: This checksum calculation is actually a generic IP functionality, maybe bring it to the core?

PR #163, opened 2020-07-20
NetworkControl: renew ip only if there is no leased IP available

HaseenaSainul:
Currently every Network.1.request/reload call initiate for either DHCP request or DHCP Discover even there is an active ip(unleaded ip). In this case, if try for ip renewal, then DHCP server will reply with NAK or offer new IP, since this is already under used list. Hence the the IP address will be changed to new IP with every request.

Actually no need to call DHCP request/Discover if the interface is up and running with valid ip (ip is not un leased). So added check to initiate IP request there is no leased IP. Along with that lease time expire logic also added to renew the IP once the IP is unleased.


NetworkControl/DHCPClientImplementation.h
```diff
@@ -731,6 +737,11 @@ namespace Plugin {
         {
             return _leasedOffer;
         }
+        inline void ClearLeasedOffer()
+        {
+             Offer offer;
+             _leasedOffer = offer;
```
mipo57, line 743: Very small remark - I think adding Offer::Clear() or Offer::Reset() function should be nicer :)

NetworkControl/NetworkControl.h
```diff
@@ -357,23 +359,30 @@ namespace Plugin {
...
+                }
             }
 
             void GetIP(const Core::NodeId& preferred)
             {
                 auto offerIterator = _client.UnleasedOffers();
                 if (offerIterator.Next() == true) {
                     Request(offerIterator.Current());
-                } else {
+                } else if (_client.LeasedOffer().IsValid() != true) {
```
mipo57, line 373: Not sure why `_client.LeasedOffer().IsValid() != true` requirement is only added in case where no valid offers are present. Shouldn't it also be present in Reqest() path?

NetworkControl/DHCPClientImplementation.h
```diff
@@ -793,8 +815,10 @@ namespace Plugin {
             _adminLock.Lock();
 
             _leasedOffer = offer;
-            _unleasedOffers.remove_if([offer] (Offer& o) {return o.Id() == offer.Id();}); 
-            
+            _unleasedOffers.remove_if([offer] (Offer& o) {return o.Id() == offer.Id();});
+
+            _activity.Revoke();
+            _activity.Schedule(Core::Time::Now().Add(_interval * 1000));
```
mipo57, line 821: Probably there is a reason, but I just wonder - why would we like to check if the lease interval has ended every n-th second, rather than just schedule one-time event after leaseTime?

NetworkControl/DHCPClientImplementation.h
```diff
@@ -944,6 +969,20 @@ namespace Plugin {
...
+        friend Core::ThreadPool::JobType<DHCPClientImplementation&>;
+        void Dispatch()
+        {
+            _adminLock.Lock();
+            uint32_t leasedTime = _leasedOffer.LeaseTime();
+            if (leasedTime > 0) {
+                _leasedOffer.LeaseTime(leasedTime - _interval);
+                _activity.Schedule(Core::Time::Now().Add(_interval * 1000));
+            } else {
+                _leaseExpiredCallback(_leasedOffer);
```
mipo57, line 981: This is a very last-moment way of extending the lease. If we do it this way, after expire we are essentially requesting a new IP rather than extending the current lease. Moreover, by calling ClearIPs() we will lose IP for a short period of time - that could unnecessarily negatively impact the system. 
A better way to do it is to renew the lease after T2 = C * LeaseTime, where C < 1 (I think C = 0.85 + e but I might be wrong).  Also then we should not clear IP, but rather call Request() on current leased one. This way IP will always be assigned and there will be no disruption. 

There is a very nice ASCII-art graphic at DHCP specification https://tools.ietf.org/html/rfc2131#page-34. Essentially it boils down to calling Request on currently leased offer after 0.85 * LeaseTime. As you can see from the diagram, there is also a T1 path with C=0.5. However, this path would be more tricky, as we would have to implement the non-broadcast version REQUEST

NetworkControl/DHCPClientImplementation.h
HaseenaSainul: Thanks @mipo57 for the comments. will address

NetworkControl/NetworkControl.h
HaseenaSainul: yea, it is applicable to both cases, but if unleaded is already there, I preferred to go with that

NetworkControl/DHCPClientImplementation.h
HaseenaSainul: Just to ensure not to miss the renewal

NetworkControl/DHCPClientImplementation.h
HaseenaSainul: Thanks Michal, 
Yea, I think we have to take care only T1 case, if the T1 case sendup with failure, then our logic will request for the rebinding/discover. so will handle the renewal based on the expected T1 value

NetworkControl/NetworkControl.h
```diff
@@ -392,8 +403,16 @@ namespace Plugin {
                 _parent.event_connectionchange(_client.Interface().c_str(), offer.Address().HostAddress().c_str(), status);
             }
 
-            inline void Request(const DHCPClientImplementation::Offer& offer) {
+            void LeaseExpired(const DHCPClientImplementation::Offer& offer) {
+                MakeUnleased();
```
mipo57, line 407: If I see correctly MakeUnleased() will put currently leased offer at the end of the unleashed list. Then wen we call GetIP which will take the first offer on the list. So if unleashed list is not empty, we will request for another Offer (and another IP) instead of renewing current lease. I think GetIP would need to be modified to use unlesedlist in a stack-like manner, or just implement Renew method in dhcp, that would just call Request on currently leased offer :)

HaseenaSainul: Thanks @mipo57, good point and it is applicable in other cases as well, i.e, need to keep the new offer or leased offer at the front of the list. But in the current logic, we are not maintaining more than one entry in the list, hence this issue will not come. If we get a DHCP NAK or unresponsive offer from potential candidates, then it will be removed from the list. So at the end there will be only one entry in the unleaded offer list.

mipo57: I think we can have more entries in a list. If we send the DISCOVER packet, we will get responses from every DHCP server in the network. In the case of a single server that will work, however, if there are more than 1 server, they can send an offer before whole DHCP transaction is made, so we will end up with 2 offers in a list. So if the first one happens to be ok, then the second one will be left on that list.
Of course, there is a discussion if we really need to keep the second offer if the first one happens to be ok :). However, at the moment - correct me if I'm wrong - it's not done so it can lead to IP switch on extending the lease

HaseenaSainul: yes, if it failed to renew, then it will try with discover cycle

HaseenaSainul: Adding the leased and new discover packet to the start of the unleaded list.

PR #169, opened 2020-07-24
HP40A-217:Fix for Connect hang issue

HaseenaSainul:
1. Avoid Lock over ConnectRequest calls to avoid lock over Trigger() which end up with deadlock
2. Documentation in both WifiControlJsonRPC API description and WifiControl.md files are updated

### DIFF
--- a/WifiControl/Controller.h
+++ b/WifiControl/Controller.h
@@ -851,16 +851,15 @@ namespace WPASupplicant {
                 _adminLock.Lock();
 
                 if (callback == _callback) {
-                    _callback = nullptr;
                     _adminLock.Unlock();
 
                     _parent.Revoke(this);
 
+                    _adminLock.Lock();
+                    _callback = nullptr;
                     result = Core::ERROR_NONE;
                 }
-                else {
-                    _adminLock.Unlock();
-                }
+                _adminLock.Unlock();
 
                 return (result);
             }
@@ -890,18 +889,18 @@ namespace WPASupplicant {
                 else {
                     _state = connection::WAITING;
                 }
-                    
+                _adminLock.Unlock();
+
                 if ((newCommand.empty() == false) && (Request::Set(newCommand) == true)) {
                     _parent.Submit(this);
                 } 
-                else if (_state != connection::WAITING) {
-                
-
-                    if (_callback != nullptr) {
+                else {
+                    _adminLock.Lock();
+                    if ((_state != connection::WAITING) && (_callback != nullptr)) {
                         _callback->Completed(result);
                     }
+                    _adminLock.Unlock();
                 }
-                _adminLock.Unlock();
             }
             void Completed(const uint32_t result) {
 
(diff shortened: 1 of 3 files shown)


WifiControl/Controller.h
```diff
@@ -802,31 +802,43 @@ namespace WPASupplicant {
         public:
```
MFransen69, line 802: Remark I cannot put the right location:

also there is one thing still scary to me (has nothing todo with your changes in itself) it is the Revoke being outside the lock here: https://github.com/rdkcentral/ThunderNanoServices/blob/71d9ec6aee48b4f9aa6a7937e837daf9bfe5709d/WifiControl/Controller.h#L857. The callback has already been set to null meaning a new connect request could be submitted if we just have a context switch between the Unlock and the Revoke here (perhaps that was what resulted in the locks in the Controller over the Invoke and Revoke calls?).
If we put the Revoke here also inside the lock that is a little risky as if some other thread would have the Controller lock and then ends up calling Connect at the moment we have the lock here and go to the Controller::Revoke call we have a deadlock again. Best solution to prevent this scenario is to only set the _callstack to null after the revoke has been completed.

WifiControl/Controller.h
```diff
@@ -802,31 +802,43 @@ namespace WPASupplicant {
...
             uint32_t Invoke(IConnectCallback* callback, const string& ssid, const uint64_t& bssid) {
 
-                uint32_t result = (_callback != nullptr ? Core::ERROR_INPROGRESS        : 
-                                  (bssid     == 0       ? Core::ERROR_INCOMPLETE_CONFIG : 
-                                  (_parent.Current().empty() == false ? Core::ERROR_ALREADY_CONNECTED :
-                                                          Core::ERROR_UNKNOWN_KEY       )));
+                uint32_t result = ((bssid == 0) ? Core::ERROR_INCOMPLETE_CONFIG :
+                                  ((_parent.Current().empty() == false) ? Core::ERROR_ALREADY_CONNECTED :
```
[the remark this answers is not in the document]
HaseenaSainul: Thank you marcel for the comments, Agree on this

WifiControl/Controller.h
```diff
@@ -883,13 +896,12 @@ namespace WPASupplicant {
                 }
```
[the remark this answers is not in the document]
HaseenaSainul: yes, it will also end up a another deadlock

WifiControl/Controller.h
HaseenaSainul: Agree on the suggestion

WifiControl/Controller.h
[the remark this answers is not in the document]
HaseenaSainul: Updated the changes

WifiControl/Controller.h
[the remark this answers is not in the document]
HaseenaSainul: updated

WifiControl/Controller.h
HaseenaSainul: Updated

WifiControl/Controller.h
```diff
@@ -890,18 +889,18 @@ namespace WPASupplicant {
                 else {
                     _state = connection::WAITING;
                 }
-                    
+                _adminLock.Unlock();
+
                 if ((newCommand.empty() == false) && (Request::Set(newCommand) == true)) {
```
MFransen69, line 894: Is there a reason not to Unlock it inside this if? If it is not entering the if doing the Unlock and Lock right after each other is rather expensive (and can lead to synchronization issues but that I did not check)

HaseenaSainul: Updated

PR #173, opened 2020-07-28
[DisplayInfo] Nexus, use HDMI resolution instead of graphics resoluti…

sebaszm:
…on, limit number of notifications

### DIFF
--- a/DisplayInfo/Nexus/PlatformImplementation.cpp
+++ b/DisplayInfo/Nexus/PlatformImplementation.cpp
@@ -46,7 +46,7 @@ public:
        , _height(0)
        , _verticalFreq(0)
        , _connected(false)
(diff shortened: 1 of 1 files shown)


DisplayInfo/Nexus/PlatformImplementation.cpp
```diff
@@ -46,7 +46,7 @@ class DisplayInfoImplementation : public Exchange::IGraphicsProperties, public E
        , _height(0)
        , _verticalFreq(0)
        , _connected(false)
-       , _hdcpprotection(HDCPProtectionType::HDCP_Unencrypted)
```
MFransen69, line 49: Why did you remove the enum type here, was it causing trouble? Indeed not strictly necessary but it is allowed and provides some more protection for a collision than without

PR #175, opened 2020-07-29
Modified to SET hex string for ssid so as to handle unicode characters within ssid

neeraj-deshpande:


WifiControl/Network.h
```diff
@@ -351,6 +351,22 @@ namespace WPASupplicant {
...
+                    result += input.substr(pos + 2, 2);
+                    pos += 3;
+                } else {
+                    result += hexdigits[input[pos] >> 4];
+                    result += hexdigits[input[pos] & 15];
+                }
+            }
+            return result;
+        }
+
```
HaseenaSainul, line 369: Can we use the Core::ToHexString() for this conversion ?

neeraj-deshpande: Core::ToHexString() does not convert unicode character present in the input string e.g. Neeraj\xe2\x80\x99s iPhone (or Neeraj’s iPhone). Will check if Core::ToHexString() can be modified to do that.

PR #176, opened 2020-07-29
[NetworkControl] Moved IPFrame and UDPFrame to core

mipo57:
Moved some IP & UDP frame logic to core.
@pwielders not sure if that is what you meat, that's why I do a PR on a branch rather than commit


MFransen69:
Not going to master but to other dev branch. I'll create a correct pull request

PR #192, opened 2020-08-21
[examples] Add a stateController plugin for testing Bartjens Law!

pwielders:


krp97 (dismissed):
I've added the `PLUGIN_STATECONTROLLER` option at `examples/CMakeLists.txt`, because that was missing for a linux build.

krp97 (approved):

PR #193, opened 2020-08-23
[VOLUMECONTROL] Add traces so we can debug the volume control.

pwielders:


Damian-Danylko (approved):

Damian-Danylko:
Looks good for me.

VolumeControl/VolumeControlImplementation.cpp
```diff
@@ -59,23 +59,27 @@ namespace Plugin {
 
     uint32_t VolumeControlImplementation::Muted(const bool muted)
     {
+        TRACE(Trace::Information, (_T("Set Muted: %s"), muted ? _T("true") : _T("false")));
         return _platform->Muted(muted);
```
MFransen69, line 63: Apparently the setter returns a error code, or is it always 0 by design? In case it isn't you might want to include it in the trace as well

pwielders: The error code is already printed in the platform layer.

pwielders: https://github.com/rdkcentral/ThunderNanoServices/pull/193/files#diff-eae5d6c96581cde7ea0ef6fff6b9e688R61 Same as in the Volume() setup

pwielders: The goal of this exercise was  to whow that the JSONRPC was called (this is the first location where we can insert code for JSONRPC, this interface is using the new feature JSON from COMRPC). SO this line reports that the JSONRPC reuqetsed it, Than it goes down to the Nexus Layers and there "seems" to be an issue there so there it prints the error code and what it sends down to the hardware: https://github.com/rdkcentral/ThunderNanoServices/pull/193/files#diff-eae5d6c96581cde7ea0ef6fff6b9e688R61

PR #200, opened 2020-08-28
DIALServer: Added logic to retry for the configured interface ready

HaseenaSainul:
In the case of both eth0 and wlan0 are connected to device and wlan0 as the configure DIAL interface, DIAL is not activating during the WPEFramework boot up. 

During the analysis found that Wlan0 is getting activated only after some times the eth0 ready and In case of DIAL, it should try to activate once it get a Network event ready (event can be form any interface, but here it is always form eth0). So in such case DIAL will not get configured interface and the initialisation will be failed. So added some retry mechanism at the boot up to handle this situation.

### DIFF
--- a/DIALServer/DIALServer.cpp
+++ b/DIALServer/DIALServer.cpp
@@ -41,7 +41,8 @@ namespace Plugin {
 
     static Core::ProxyPoolType<Web::TextBody> _textBodies(5);
 
-    static constexpr uint16_t TimeToInterfaceReady = 1000;
+    static constexpr uint8_t InterfaceRetryCount = 5;
+    static constexpr uint16_t InterfaceWaitTime = 1000;
 
     /* static */ const Core::NodeId DIALServer::DIALServerImpl::DialServerInterface(_T("239.255.255.250"), 1900);
     /* static */ std::map<string, DIALServer::IApplicationFactory*> DIALServer::AppInformation::_applicationFactory;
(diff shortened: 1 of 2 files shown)


pwielders:
The idea of the subsytems is that they will only be made "Active" if the subsystem they present is really available. So if the Network SubSystem is made available it means that there is at least a possibility for the system to use the network. Which means that at least 1 interface has a valid IP address.
1) If there are multiple interfaces and the middleware (UI) want the DIALServer to be "only" available on the wlan interface, it is up to the middleware to enable the DIALServer after they have detected that this specific interface is available. 
Or
2) The Network SubSystem is only triggered if "all" network interfaces are available.

I think we should not start retrying until that interface becomes available since this leads to undeterministic and misleading behavior. The Activate of the DIALServer, might suggest all is oke, but at the end, it will only happen if the WLAN interface becomes available (when will this happen?), which might trigger a new series of tickets.

If we do it here, should we also start retrying to get an interface on other plugins like the locationSync? If so would it not be better to remove the Network Event than ? Cause evenm if it signals Oke, we still need to  code for retrying ;-)

So I think we should resolve this issue at the root-casue (network subsystem or middleware)

HaseenaSainul:
Agree Pierre on your comments. 
So as per your suggestion this has to be handled outside the DIALServer, so I will close this PR right now.

PR #204, opened 2020-09-01
Development/streamer fixes

krp97:
This commit introduces two fixes for the Streamer plugin:
1. Missing `INTERFACE_AGGREGATE` call for `IPlayer` interface.
2. Implemented extended IStream interface as per rdkcentral/Thunder#249


Streamer/Frontend.h
```diff
@@ -355,6 +354,33 @@ namespace Player {
...
+                _adminLock.Unlock();
+            }
+            uint32_t Detach(IStream::ICallback* callback) override
+            {
+                uint32_t result = Core::ERROR_NONE;
+                
+                _adminLock.Lock();
+
+                if (_callback != nullptr && _callback == callback) {
+                    TRACE_L1("Releasing the callback");
```
pwielders, line 375: If you want to use tracing, use the TRACE macro and not the TRACE_L1

pwielders (changes requested):

pwielders (approved):

PR #207, opened 2020-09-03
Development/network cleanup

pwielders:


NetworkControl/NetworkControl.h
```diff
@@ -218,304 +304,198 @@ namespace Plugin {
...
-
-                JsonData::NetworkControl::ConnectionchangeParamsData::StatusType status;
-                if (result == true) {
-                    _parent.RequestAccepted(_client.Interface(), offer);
-                    status = JsonData::NetworkControl::ConnectionchangeParamsData::StatusType::CONNECTED;
-                } else {
-                    _parent.RequestFailed(_client.Interface(), offer);
-                    status = JsonData::NetworkControl::ConnectionchangeParamsData::StatusType::CONNECTIONFAILED;
+                else if ( (_client.Lease() == _offers.front()) && (_retries++ < _maxRetries) ) {
+                    // Looks like the acknwledge did not get a reply, should we retry ?
```
HaseenaSainul, line 409: typo error in comments 'acknwledge'

NetworkControl/NetworkControl.cpp
```diff
@@ -473,21 +473,27 @@ namespace Plugin
...
 
-                DHCPClientImplementation::Offer::JSON lease;
+                Store storage;
                 Core::OptionalType<Core::JSON::Error> error;
-                if (lease.IElement::FromFile(leaseFile, error) == true) {
-
-                    _client.AddUnleasedOfferToEnd(lease.Get());
+                if (storage.IElement::FromFile(leaseFile, error) == true) {
+                    
+                    Core::JSON::ArrayType<Entry>::Iterator loop(storage.Elements());
```
HaseenaSainul, line 490: Trying to keep Entry objects on Settings, since both are different.

NetworkControl/NetworkControl.cpp
```diff
@@ -152,26 +155,18 @@ namespace Plugin
...
-                        std::make_tuple(Core::ProxyType<DHCPEngine>::Create(this, interfaceName, _persistentStoragePath)));
-                    _interfaces.emplace(std::piecewise_construct,
-                        std::make_tuple(interfaceName),
-                        std::make_tuple(index.Current()));
+                    _dhcpInterfaces.emplace(std::piecewise_construct,
+                        std::forward_as_tuple(interfaceName),
+                        std::forward_as_tuple(*this, interfaceName, _responseTime, _retries));
 
                     JsonData::NetworkControl::NetworkData::ModeType how(index.Current().Mode);
                     if (how == JsonData::NetworkControl::NetworkData::ModeType::MANUAL) {
```
HaseenaSainul, line 165: Not keeping the Interface mode anywhere and trying to use it later on (example from RefreshDNS()

NetworkControl/NetworkControl.cpp
```diff
@@ -825,13 +686,39 @@ namespace Plugin
...
+                servers.emplace_back(*pointer);
                 pointer++;
             }
 
+            // Strawl through the list of dhcp entries and get the DNS entries from there
+            std::map<const string, DHCPEngine>::iterator loop(_dhcpInterfaces.begin());
+
+            while (loop != _dhcpInterfaces.end()) {
+                
+                if (loop->second.Info().Mode() == JsonData::NetworkControl::NetworkData::ModeType::DYNAMIC) {
```
HaseenaSainul, line 702: Trying to use non stored value, hence this will fail to retrieve the DNS info received from the DHCP server

HaseenaSainul (changes requested):

NetworkControl/NetworkControl.cpp
```diff
@@ -508,30 +514,25 @@ namespace Plugin
...
+                if (interface.IsValid() &&
+                    interface.HasMAC()  &&
+                    interface.IsRunning()) {
 
                     uint8_t mac[6];
                     interface.MACAddress(mac, sizeof(mac));
-                    entry->second->UpdateMAC(mac, sizeof(mac));
-                    entry->second->GetIP();
+                    index->second.UpdateMAC(mac, sizeof(mac));
+                    index->second.Discover(index->second.Info().Address());
```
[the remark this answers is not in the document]
pwielders: This is part of the method Reload, this is only triggered if a REST call/JSONRPC call is coming trigegr this functionality, or during initialization, if this interface is configured for DHCP.
If your remark is, can we avoid during initialization a DHCP request for this interface, we can not, sonce we do not know what the lease period for this interface was, or is. To know the lease time, the Network control *must* execute a discovery.
Maybe we could make logic to check if the IP in there is the IP also found in the persistent store and if so, start with an Ack sequence, as an optimization. But lets do this optimization in a second run.

NetworkControl/NetworkControl.cpp
```diff
@@ -540,183 +541,62 @@ namespace Plugin
...
+            const string name (adapter.Name());
+            std::map<const string, DHCPEngine>::iterator index (_dhcpInterfaces.find(name));
+            
+            if (index != _dhcpInterfaces.end()) {
+                bool hasValidIP = ExternallyAccessible(adapter);
+                if (std::find(_requiredSet.cbegin(), _requiredSet.cend(), name) != _requiredSet.cend()) {
+                    count++;
+                    fullSet &= hasValidIP;
                 }
+                validIP |= hasValidIP;
```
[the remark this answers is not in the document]
pwielders: If there are no interfaces set in the required list, the network subsystem can be activated if any interface has a validIP address so in case the required list is empty, the subsystem is based on this falg. But I saw already that the setting of the Network SubSystem is not correct, so that it is not used might be a reason :-)

NetworkControl/NetworkControl.cpp
```diff
@@ -540,183 +541,62 @@ namespace Plugin
...
-    /* virtual */ uint32_t NetworkControl::RemoveDNS(IIPNetwork::IDNSServers * dnsEntries)
-    {
-        uint32_t result = Core::ERROR_UNKNOWN_KEY;
-        bool syncDNS = false;
+        ASSERT (count <= _requiredSet.size());
 
-        dnsEntries->Reset();
-
-        _adminLock.Lock();
+        fullSet &= (count == _requiredSet.size());
```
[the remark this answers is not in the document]
pwielders: As stated above, I saw that there is something wrong in detecting the SubSystem logic. I need to revisit it. Thanks for this remark, casue it might well be the issue :-)

NetworkControl/DHCPClientImplementation.h
```diff
@@ -590,229 +449,162 @@ namespace Plugin {
...
+
             _state = IDLE;
+
             // Do not wait for UDP closure, this is also called on the Communication Thread !!!
-            if (SocketDatagram::IsOpen())
+            if (SocketDatagram::IsOpen()) {
                 SocketDatagram::Close(0);
+            }
+
             _adminLock.Unlock();
```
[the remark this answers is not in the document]
pwielders: With what purpose? It will speed up the code sligtly but make the reading of it a bit more difficult. It would be a trade off between speed and readability. This is executed very rarely, but we could

NetworkControl/DHCPClientImplementation.cpp
```diff
@@ -203,49 +47,44 @@ namespace Plugin {
...
 
-            DHCPIPPacket frame(dataFrame, maxSendSize, true);
-            result = Message(frame.DHCPFrame(), frame.MaxDHCPFrameSize());
-
-            if (result > 0) {
-
-                frame.DHCPFrameSize(result);
-                result = frame.TotalSize();
+            if ( (_modus == CLASSIFICATION_DISCOVER) || (_offer.Source().IsValid() == false) ) {
+                RemoteNode(BroadcastServerNode);
```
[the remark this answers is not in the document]
pwielders: This is still work in progress,  as you can see there is still some commented out stuff, I think now it is already different,

NetworkControl/NetworkControl.cpp
pwielders: ???

HaseenaSainul: Second field of map::info is Settings and we are trying to keep the type Entry type on thia (loop is type of Entry)

HaseenaSainul (approved):

PR #216, opened 2020-09-09
Modified to fetch & build brcm specific code from external repository

neeraj-deshpande:


MFransen69 (dismissed):

CMakeLists.txt
```diff
@@ -22,6 +22,8 @@ find_package(WPEFramework)
 # are located in the cmake directory. Include it in the search.
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/")
 
+option(CODE_FROM_EXTERNAL_REPOSITORY "Include code from External Repository" ON)
```
bramoosterhuis, line 25: I would expect to switch this new feature default off and if everything is fine after a while switch it on.

MFransen69 (changes requested):
Hi Neeraj, think I was not clear in my last request, sorry. It is not a special flag to override when to take the sources from Gerrit, the idea was that you could specify for a component to not take the sources from Gerrit (or RDK Central) but you could specify a folder to get the sources from (so you can have them local on a disk or something similar instead)

MFransen69 (approved):

MFransen69 (approved):

PR #229, opened 2020-09-21
Performance Monitor: Thunder Performance flag enabled added + doc updates

HaseenaSainul:


pwielders (commented):
Think this does not work. Try building core without PERFORMANCE and this with, you probably en up getting all 0's as result, I think

HaseenaSainul:
Since this functionality is only work once the PERFORMACE flag is enabled for core. So I thought is to enable this plugin only if the PERFORMANCE_TEST is enabled for core (from the OE or BR build make files). Thats why added this flag directly

PerformanceMonitor/CMakeLists.txt
```diff
@@ -37,6 +37,10 @@ target_link_libraries(${MODULE_NAME}
         ${NAMESPACE}Plugins::${NAMESPACE}Plugins
         ${NAMESPACE}Definitions::${NAMESPACE}Definitions)
 
+target_compile_definitions(${MODULE_NAME}
```
pwielders, line 40: I think we can take this part out now. We added it through the Thunder layer. I guess part as a plugins variable.

HaseenaSainul: No @pwielders, since the PerformanceAdministrator is defined inside only Request.h and its not part of any libs, so still we need this flag to get the Thunder_PERFORMANCE as 1 from the plugin

HaseenaSainul: Yes removed this now, since this is exposed through compile settings

pwielders (approved):

PR #235, opened 2020-09-28
[DAILServer] Add Amazon Prime configuration

bramoosterhuis:


artur-gebicz-ml (approved):
LGTM

pwielders (approved):
This is oke for MCA Explora, for MCA Streame, the whole setting passive will not be taken into account as the passive active switch will be based on the CallSign again.

PR #240, opened 2020-10-03
[DISPLAYINFO] Get width/height in cm from EDID

pawel-stanek-ml:
+ small fixes


pawel-stanek-ml:
Note this is again development/InfoPlugins which is not quite aligned with current master. It depsnds on https://github.com/rdkcentral/Thunder/pull/322

PR #243, opened 2020-10-06
WifiControl: Revoke Autoconnect before releasing _controller

HaseenaSainul:


WifiControl/WifiControl.cpp
```diff
@@ -139,16 +139,15 @@ namespace Plugin
     /* virtual */ void WifiControl::Deinitialize(PluginHost::IShell * service)
     {
 #ifndef USE_WIFI_HAL
+
```
pwielders, line 142: Should the _autconnect.Revoke, not be positioned two line up, before the #ifndef USE_WIFI_HAL as i think it is always applicable.

HaseenaSainul: Since we are using auto connect only ifndef USE_WIFI_HAL case, thats why kept it inside this

HaseenaSainul: I think this code will not compile properly for USE_WIFI_HAL case, since we did not use the preprocessor check properly

pwielders (approved):

PR #245, opened 2020-10-06
Development/dial refactor

bramoosterhuis:


DIALServer/Netflix.cpp
```diff
@@ -25,210 +25,178 @@
...
-                TCHAR* encodedPayload = reinterpret_cast<TCHAR*>(ALLOCA(maxEncodeSize));
-                Core::URL::Encode(payload.c_str(), static_cast<uint16_t>(payload.length()), encodedPayload, maxEncodeSize);
-                query = query + _T("&dial=") + encodedPayload;
-            }
+    static string Query(const string& params, const string& payload)
+    {
+        string query = params;
+#ifdef NETFLIX_VERSION_5_1
+            // Set proper launch type, i.e. launched by DIAL
+            query += _T("&source_type=12");
```
pwielders, line 33: Lets make this a field in the config of the Plugin::DIALServer::Config::App& config, so it is coming from the configuration, in stead of hard codinf. Than also the preprocesspr stuff can dissaper here. So add a field "identification": "source_type=12" or "identification": "iid=7637f789". This way we move the netflix version to the config, no need to rebuild fr any netflix version.

DIALServer/CMakeLists.txt
```diff
@@ -45,18 +45,17 @@ add_library(${MODULE_NAME} SHARED
...
-if(PLUGIN_DIALSERVER_ENABLE_YOUTUBE)
+if(PLUGIN_DIALSERVER_ENABLE_YOUTUBE AND "${PLUGIN_DIALSERVER_YOUTUBE_MODE}" STREQUAL "active")
     target_sources(${MODULE_NAME}
         PRIVATE
             YouTube.cpp)
 endif()
 
-if(PLUGIN_DIALSERVER_ENABLE_NETFLIX)
-    target_sources(${MODULE_NAME}
+if(PLUGIN_DIALSERVER_ENABLE_NETFLIX AND "${PLUGIN_DIALSERVER_NETFLIX_MODE}" STREQUAL "active")
```
pwielders, line 54: As the name of the handler might, in the future or now already" deviate from the callsign suggest to use something like PLUGIN_DIALSERVER_NETFLIX_HANDLER. If you set it you mean I want active mode and you can potentially define another handler name than the callsign.

pwielders (approved):

PR #249, opened 2020-10-08
Hello World Plugin example

sramani-metro:
First Commit


MFransen69:
As disused with Santhosh we will not merge this PR

PR #260, opened 2020-10-27
VolumeControl:AMLogic: revert formula changes

HaseenaSainul:


modeveci (approved):
This change will be handled by audio driver. They will provide a new interface which will take percentage.

PR #267, opened 2020-11-02
[COBALT] Adding the language and the connection properties for the Co…

pwielders:
…balt certification.


modeveci:
It looks good to me, I will give a try

modeveci (approved):

PR #272, opened 2020-11-10
METROL-169: fix to renew IP while switching the subnet or routers

HaseenaSainul:


pwielders (approved):
Good fix! Good analyses !

PR #273, opened 2020-11-11
Modified Cobalt plugin to rename debugport config parameter as  inspector.

neeraj-deshpande:
This is done for uniformity as well as inspector param is also understood by ThunderUI.


pwielders (commented):
Why this change? Name change oke, but now you are usggestion that the inspector port can be bound to an IP and thus interface, but you are not using it, so it will not work. Or call it inspector port, or also apply the IP. Now you are giving the wrong impression to the user of this plugin.

pwielders (approved):

PR #275, opened 2020-11-17
Added AMLogic implementation

krp97:
This commit introduces a reworked version of the DisplayInfo
implementation for an AMLogic hp40a platform. The notifications
about a state change of the connection are handled via
incoming udev messages.

TODO:
Currently, the udev state change messages are not being generated
on a succesfull authentication. The lack of udev messages
is being investigated. Tracking link:
https://jira.rdkcentral.com/jira/browse/HP40A-603


DisplayInfo/DisplayInfo.cpp
```diff
@@ -84,16 +87,17 @@ namespace Plugin {
             _graphicsProperties = nullptr;
         }
 
+        ASSERT(_hdrProperties != nullptr);
```
pwielders, line 90: The RaspberryPi has no HDR properties and thus no implementation for this interface. It should still work, but this might kill it.

DisplayInfo/DisplayInfo.cpp
```diff
@@ -33,14 +33,17 @@ namespace Plugin {
...
         string message;
-        Config config;
-
-        config.FromString(service->ConfigLine());
         _skipURL = static_cast<uint8_t>(service->WebPrefix().length());
 
         _connectionProperties = service->Root<Exchange::IConnectionProperties>(_connectionId, 2000, _T("DisplayInfoImplementation"));
         if (_connectionProperties != nullptr) {
 
+            Exchange::IWebDriver* configConnection = _connectionProperties->QueryInterface<Exchange::IWebDriver>();
```
pwielders, line 41: Agree that you need an interface to configure some implementations, and this is good, but reusing an interface for the WebDriver is a bit debatable. Can you make a Generic IConfiguration interface, so that it is specific and easier to understand.

pwielders (changes requested):
Most remarks are minor, but the IWebdriver to configure really requires an update :-)

pwielders:
@mhughesacn Seems we have a black duck issue again :-) Can you guide us ?

mhughesacn:
Hi @pwielders - This was sorted yesterday and the check cleared.  (The problem was a misconfiguration in BlackDuck which I fixed as soon as I saw it.)

pwielders:
@mhughesacn Thanks !

pwielders (approved):

PR #311, opened 2020-12-16
DisplayInfo: fix to avoid exception due to empty values

HaseenaSainul:


DisplayInfo/Amlogic/PlatformImplementation.cpp
```diff
@@ -209,9 +209,13 @@ namespace Plugin {
         uint64_t GetMemory(const std::string& key) const
         {
             auto extractNumbers = [](const std::string& str) {
-                auto first = str.find_first_of("0123456789");
-                auto last = str.find_last_of("0123456789");
-                return std::stoul(str.substr(first, last - first + 1));
+                string value;
+                if (str.empty() != true) {
```
krp97, line 213: For consistency, could you make this line: `if(!str.empty())`?

krp97 (changes requested):

DisplayInfo/Amlogic/PlatformImplementation.cpp
HaseenaSainul: Done

krp97 (approved):

PR #347, opened 2021-01-27
PlayerInfo:Dolby:Amlogic: Trace compile fix

HaseenaSainul:


PlayerInfo/Dolby/AMLogic/Implementation.cpp
```diff
@@ -23,6 +23,7 @@
 #include <string.h>
 #include "hardware/audio.h"
 
+using namespace WPEFramework;
```
MFransen69, line 26: Hi Haseena, if I may ask why did you add this? Was there something else not compiling? For the change below it would not be needed it I'm right

HaseenaSainul: Yes it is required to compile the Trace log (Trace Category is under WPEFramework namespace)

modeveci (dismissed):
Worked on yocto build

pwielders:
Could we also make this a dedicated fix, e.g. i.s.o Trace::Information -> WPEFramework::Trace::Information. or the otherway around, the line just above this trace, remove the WPEFramework namespace before the ERROR_<CODe>.
Currently (for @modeveci  to learn dutch saysing) "hinkt op twee gedachtes" (it is using both possible ways, and declering it global, which if possible, should be avoided if it is not used a lot, and some statement are fully qualified)

HaseenaSainul:
Updated as per @pwielders comments

PlayerInfo/Dolby/AMLogic/Implementation.cpp
MFransen69: Eh, but now you removed it, so it was not needed then?

modeveci: "Hinkt op twee gedachtes" :)

HaseenaSainul: yea, instead used it directly from the Trace API WPEFramework::Trace::Information

modeveci (approved):

pwielders (approved):

PR #349, opened 2021-01-28
Fix DIAL additionaldataUrl port settings

Damian-Danylko:


DIALServer/DIALServer.cpp
```diff
@@ -286,7 +286,6 @@ namespace Plugin {
             _dialURL = Core::URL(service->Accessor());
             _dialURL.Host(selectedNode.HostAddress());
             _dialPath = '/' + _dialURL.Path().Value();
-            _webServerPort = _dialURL.Port();
```
pwielders, line 289: Tgis looks scary, now the _webServerPort is not set. PLease double check if this is really good.

DIALServer/DIALServer.cpp
```diff
@@ -380,7 +379,7 @@ namespace Plugin {
         } else {
             // FIXME: At the moment part of additionalDataUrl parameter is hardcoded, localhost is obligatory by Netflix 
             // but rest of the path can be created dynamically or should be retrived from configuration    
-            const string additionalDataUrl = (_T("http://localhost") + ((_webServerPort != 80)? _T("") : _T(":") + Core::NumberType<uint16_t>(_webServerPort).Text()) + _T("/Service/DIALServer/Apps/") + app.Name() + _T("/") + _DefaultDataExtension);
+            const string additionalDataUrl = (_T("http://localhost") + ((_webServerPort == 80)? _T("") : _T(":") + Core::NumberType<uint16_t>(_webServerPort).Text()) + _T("/Service/DIALServer/Apps/") + app.Name() + _T("/") + _DefaultDataExtension);
```
pwielders, line 382: This I understand and indeed is a bug. Good Fix

pwielders (changes requested):
Please double check the removal of the _webServerPort. I think it is an accidental remove and it should be in

DIALServer/DIALServer.cpp
Damian-Danylko: I forgot to commint header file. Now implementation should cover both case when WebServer is off and on. In both cases 
_webServerPort will be set properly.

pwielders (approved):

PR #352, opened 2021-02-09
[ResourceMonitor] Refactored plugin 

lukiwa:
ResourceMonitor is now able to monitor processes individually (WPEFramework, WPEProcess, WPENetworkProcess, etc) and all together (select name "WPE" for monitoring all WPE related processes) and collects information about memory usage, and puts it into a .csv file. Switched to RAII containers where possible. Changed GCC __builtin_popcount function to STL equivalent. Tested on RPI and inside a docker container.


ResourceMonitor/ResourceMonitorImplementation.cpp
```diff
@@ -1,503 +1,336 @@
...
+
+            void Dispatch()
+            {
+                Collect();
             }
 
-            StartLogLine(processIds.size());
-            for (std::pair<Core::ProcessInfo, string> processDesc : processIds) {
-               Core::ProcessTree tree(processDesc.first);
+            uint32_t CountSetBits(const bool calculateUSS, const std::vector<uint32_t>& pageBuffer, const std::vector<uint32_t>& inverseMask)
```
krp97, line 208: I think we could refactor this somehow, because `CountSetBits` is essentially two functions where we trigger one or the other based on some arbitrary flag.

This:
```
 else {
   for (uint32_t index = 0; index < _bufferEntries; index++) {
     count += std::bitset < 32 > (pageBuffer[index]).count();
   }
 }
```
could be reworked into something like:
```
std::accumulate(pageBuffer.begin(), pageBuffer.end(), 0, [](int total, uint32_t page) {
    return total + std::bitset<32>(page).count();
})
```

This:
```
if (calculateUSS) {
  for (uint32_t index = 0; index < _bufferEntries; index++) {
    count += std::bitset < 32 > (pageBuffer[index] & (~inverseMask[index])).count();
  }
}
```
could potentially be replaced with:
```
[](const std::vector<uint32_t>& pageBuffer, const std::vector<uint32_t>& inverseMask){
    uint32_t count = 0;
    for (uint32_t index = 0; index < _bufferEntries; index++) {
         count += std::bitset < 32 > (pageBuffer[index] & (~inverseMask[index])).count();
    }
}
```
Especially since I think they're only used once in the `LogProcess` function. You could also just split this into two seperate functions and name them appropriately. 

Just a suggestion, let me know what you think.

ResourceMonitor/ResourceMonitorImplementation.cpp
```diff
@@ -1,503 +1,336 @@
...
 
-            _activity.Schedule(Core::Time::Now().Add(_interval * 1000));
-         }
+        private:
+            CSVFile _logfile;
+            uint32_t _memoryPageSize; //size of the device memory page
+
+            std::vector<string> _processNames; // Seen process names.
+            Core::CriticalSection _guard;
+            std::vector<uint32_t> _otherMap; // Buffer used to mark other processes pages.
```
krp97, line 246: This name could use some refactoring, `_otherMap` doesn't give the reader any insight.

krp97 (changes requested):
I'd like to discuss some minor changes, let me know what you think about them and please retest after applying all valid resolutions.

PS: Before you apply any fixes, could you align this branch with master? I think it has gone out of date.

lukiwa:
After diving deeper into the results returned by the measurements I decided to change the underlying measurement algorithm. More to be read here: https://github.com/rdkcentral/Thunder/pull/482

lukiwa:
I messed something with rebases, closing this PR.

PR #363, opened 2021-02-24
[METRO-124] Create a new LanguageAdministrator plugin

Vinodsathyaseelan:
[METRO-124] Create a new LanguageAdministrator plugin

- Reason for change :  Enhancement
    Currently languages need to be set per application, which also does not work when the application is not active. 
	We should extend the Location sync to provide the current active langue/locale (by default determined by the location) and 
    the possibility to override this from the interface (e.g. when the middle ware allows the user to select a different language).  
			   
    Also a notification of this change/event should be available so JS can respsond to this event.
	 
- Solution : 
	Instead of extending LocationSync, we have agreed to introduce a new plugin named LanguageAdministrator
    to take care of the language related requirements.


LanguageAdministrator/LanguageAdministrator.cpp
```diff
@@ -0,0 +1,186 @@
...
+    {
+        if (connection->Id() == _connectionId) {
+            ASSERT(_service != nullptr);
+            Core::IWorkerPool::Instance().Submit(PluginHost::IShell::Job::Create(_service,
+                PluginHost::IShell::DEACTIVATED,
+                PluginHost::IShell::FAILURE));
+        }
+    }
+
+    void LanguageAdministrator::StateChange(PluginHost::IShell* plugin, const string& callsign)
```
pwielders, line 99: Sorry, you where not aware of this, but yesterday a breaking change for this was added. Right now you need to split this method in 2 methods, one called void Activated (const string& callsign, PluginHost::IShell* plugin) and the other void Deactivated (const string& callsign, PluginHost::IShell* plugin).
The reason for this change is to avoid chatiness on getting the plugin->State() which might be over process boundaries and to make the reported states by Thunder less, only Activate and Deactivated are rported now..

LanguageAdministrator/LanguageAdministratorImpl.cpp
```diff
@@ -0,0 +1,102 @@
...
+        notification->AddRef();
+        _notifications.push_back(notification);
+        _adminLock.Unlock();
+    }
+
+    void LanguageAdministratorImpl::Unregister(const Exchange::ILanguageTag::INotification* notification)
+    {
+        ASSERT(notification);
+        _adminLock.Lock();
+        //TODO: Store the language code to file.
```
pwielders, line 49: old comment?

LanguageAdministrator/LanguageAdministratorImpl.cpp
```diff
@@ -0,0 +1,102 @@
...
+    void LanguageAdministratorImpl::NotifyLanguageChanged(const string& language)
+    {
+        _adminLock.Lock();
+        for (auto* notification : _notifications) {
+            notification->LanguageChanged(language);
+        }
+        _adminLock.Unlock();
+    }
+
+    /*bool LanguageAdministratorImpl::IsvalidLanguageTag(const string& language) {
```
pwielders, line 96: Can we remove dead code? The SCM will keep track of it, if you need it later on.

LanguageAdministrator/LanguageAdministratorImpl.h
```diff
@@ -0,0 +1,60 @@
...
+
+        BEGIN_INTERFACE_MAP(LanguageAdministratorImpl)
+            INTERFACE_ENTRY(Exchange::ILanguageTag)
+        END_INTERFACE_MAP
+
+        //   ILanguageTag methods
+        void Register(Exchange::ILanguageTag::INotification* observer) override;
+        void Unregister(const Exchange::ILanguageTag::INotification* observer) override;
+
+        virtual uint32_t Language(string& language /* @out */) const override;
```
pwielders, line 46: According to the coding rules and guidelines, you can only have virtual or override on a mehod, not both. Also, here you use string, which is good but in the interface specification you used std:string, which I reuested to change to string as well, than it is in line with this here..

LanguageAdministrator/LanguageAdministratorImpl.h
```diff
@@ -0,0 +1,60 @@
...
+        virtual uint32_t Language(const string& language) override;
+
+    private:
+        std::string _language;
```
pwielders, line 50: Do not use the std on string, leave it up to Thunder to select the right flavour of string unless it *must* be a char or wchar representations...

pwielders (changes requested):
Looks good a few minors and a recent change which requires a small adaption of the code. But looks good !

LanguageAdministrator/LanguageAdministratorImpl.h
Vinodsathyaseelan: Forget to remove 'virtual' after copy paste from Interface definition. I will use sting instead of std::string.

LanguageAdministrator/LanguageAdministratorImpl.cpp
Vinodsathyaseelan: Yes, old comment.  Will remove this.

Vinodsathyaseelan:
Implemented Activated() & Deactivated() functions and removed StateChange() in LanguageAdministrator plugin.
Addressed other review comments from Pierre Wielders.

pwielders (approved):

PR #367, opened 2021-03-03
cmake:  plugins are now seperate cmake projects

bramoosterhuis:


pwielders (approved):
@bramoosterhuis  take care there is one source file in here as well, Monitor.H I think it is a merge thingy!

PR #372, opened 2021-03-17
RemoteControl: LinuxDevice: Add inputDevice to the device list if it exist

HaseenaSainul:


pwielders (commented):
First off all good analyses again!!
I understand how this might lead to a crash, but looking at the code (I was assuming defensive coding was required here) I assume that the following fix would more describe the intention of the code and simplify the code :-) I guess it is only a nullptr if it was never found..

                            for (auto& device : _inputDevices) {
                                std::size_t found = deviceName.find(Core::EnumerateType<LinuxDevice::type>(device->Type()).Data());
                                if (found != std::string::npos) {
                                   ASSERT(device != nullptr);
                                    _devices.insert(std::make_pair(entry.Name(), std::make_pair(fd, device))); 
                                   break;
                                }
                            }

Without the additional check...

HaseenaSainul:
Ok thanks will update the code :)

HaseenaSainul:
@pwielders, updated the changes

pwielders:
LGTM :-)

pwielders (approved):

PR #376, opened 2021-03-25
[METROL 236] HDR luminance adjustment possibility

lukiwa:
https://github.com/rdkcentral/ThunderInterfaces/pull/36


Compositor/Compositor.h
```diff
@@ -157,6 +158,7 @@ namespace Plugin {
             Core::JSON::DecUInt32 Y;
             Core::JSON::DecUInt32 Width;
             Core::JSON::DecUInt32 Height;
+            Core::JSON::DecUInt16 Brightness;
```
pwielders, line 161: Why not define the enum? Looks like the setting (on interface) is an enum, Lets reflect that in the config

pwielders: Where is the config actually being used ? ohh this isreporting on it. Would also expect that you can set it form the config, so I ghuess it is still missing ?

pwielders (changes requested):

Compositor/Compositor.h
lukiwa: This is actually leftover from the abandoned HTTP response functionality that I planned to add, it should be deleted. You can set the brightness setting from the config in the Compositor-brcm (https://github.com/WebPlatformForEmbedded/Compositor-brcm/pull/1)

Compositor/Compositor.cpp
```diff
@@ -650,6 +650,28 @@ namespace Plugin {
         return (result);
     }
 
+    uint32_t Compositor::GetBrightness(Exchange::IComposition::Brightness& brightness) const {
+        ASSERT(_composition != nullptr);
+        const WPEFramework::Exchange::IComposition* constComposition = _composition;
```
pwielders, line 655: What is the added value of copying the pointer ? if the issue is at the executions you can also state there:
return (static_cast<const Exchange::IComposition*>(_composition)->SdrToHdrGraphicsBrightness(brightness));
but I think that is not even nessecary :-)

lukiwa: Copying is not necessary indeed, but the cast itself should remain IMO. Those are warnings I got when I remove it (and get method is calling the not-const method from the interface without them):
```
warning: ISO C++ says that these are ambiguous, even though the worst conversion for the first is better than the worst conversion for the second:
             return _composition->SdrToHdrGraphicsBrightness(brightness);
                                                                       ^
In file included from Compositor.h:24:0,
                 from Compositor.cpp:20:
../../../host/usr/arm-buildroot-linux-gnueabihf/sysroot/usr/include/WPEFramework/interfaces/IComposition.h:95:26: note: candidate 1: virtual uint32_t WPEFramework::Exchange::IComposition::SdrToHdrGraphicsBrightness(const WPEFramework::Exchange::IComposition::Brightness&)
         virtual uint32_t SdrToHdrGraphicsBrightness(const Brightness& brightness) = 0;
                          ^~~~~~~~~~~~~~~~~~~~~~~~~~
../../../host/usr/arm-buildroot-linux-gnueabihf/sysroot/usr/include/WPEFramework/interfaces/IComposition.h:94:26: note: candidate 2: virtual uint32_t WPEFramework::Exchange::IComposition::SdrToHdrGraphicsBrightness(WPEFramework::Exchange::IComposition::Brightness&) const
         virtual uint32_t SdrToHdrGraphicsBrightness(Brightness& brightness /* @out */) const = 0;
```

pwielders (approved):

lukiwa:
Changes implementation due to changes in the interface

pwielders (approved):

PR #382, opened 2021-03-26
[MODULE] With the EarlyWarning enabled it becomes even more important…

pwielders:
… to ALWAYS first include the Module.h


neeraj-deshpande (approved):
LGTM

PR #384, opened 2021-04-06
WifiControl: trim double black slash from wpa_supplicant provided SSID

HaseenaSainul:


WifiControl/Controller.h
```diff
@@ -1954,6 +1957,18 @@ namespace WPASupplicant {
             return status;
         }
 
+        inline void TrimEscapeSequence(string& str) {
+            string escapeSequence("\\\\");
```
pwielders, line 1961: Do we not ant to unescape all text going through here, like the \n, \t, \t etc..
Maybe even move an string:Unescape/string::Escape method into core, to be reused in the JSON stuff as wel ?

pwielders (approved):
More extensive testing showed that there is no way ti use (ingest) any other escapable characters rthan the backslahs and it seems that it is only applicable to the SSID. Hence why we can progress with only the unescaping of the backslash.

PR #390, opened 2021-04-12
Compositor: cmake for westeros compositor added

HaseenaSainul:


pwielders (approved):

artur-gebicz-ml (approved):
I'm not sure if should update the license text to the new one (it's still @metro 2015).
Beside that LGTM. Thanks

HaseenaSainul:
It was created during 2015 and as part cmake restructuring it was removed. Now I restored the dependencies, hence did not updated license info :)

PR #406, opened 2021-04-23
[DIALServer] Only allow chaning port of locator URL from external source

sebaszm:
Only if binding interface is set to 0.0.0.0 it is possible to change the locator host to the one from the UPNP request, in other cases (i.e from WebServer) only port is updateable.


pwielders (approved):
Less is always better :-)

PR #424, opened 2021-05-03
Svalbard: config update

HaseenaSainul:


bramoosterhuis (dismissed):

Svalbard/Svalbard.config
```diff
@@ -1,16 +1,12 @@
-set (autostart true)
+set (autostart ${PLUGIN_SVALBARD_AUTOSTART})
 set (preconditions Platform)
 
 map()
 end()
 ans(configuration)
 
 map()
-    if(PLUGIN_SVALBARD_OOP)
```
artur-gebicz-ml, line 9: I believe we should make that backward compatible, so if PLUGIN_SVALBARD_OOP then set the PLUGIN_SVALBARD_OUTOFPROCESS to true

artur-gebicz-ml (changes requested):

Svalbard/Svalbard.config
HaseenaSainul: What I understood from @pwielders is that this module is not yet used by any clients. So I think it safe to change the config now. @bramoosterhuis can you please confirm

pwielders: @artur-gebicz-ml  good point but indeed @HaseenaSainul is right here, there is no need for backwards compatibility since this has not been deployed, no need for backwards compatibility.

pwielders (approved):

PR #442, opened 2021-06-21
Fix origin for YT Protocol Compliance test cases

Damian-Danylko:


DIALServer/DIALServer.cpp
```diff
@@ -662,16 +662,23 @@ namespace Plugin {
...
-
-            if (hostPortion.IsSet() == true) {
-                Core::NodeId source(hostPortion.Value().c_str());
-
-                safe = (source.IsLocalInterface() == true) || (hostPortion.Value().find(app.Origin()) != string::npos);
+            Core::NodeId source(hostPortion.Value().c_str());
+            if (source.IsLocalInterface() == true) {
+                safe = true;
+            } else {
+                if (request.Origin.Value().find("https") == 0) { // Valid format: https://www.youtube.com, https://dial.youtube.com
```
pwielders, line 671: I think we would like to use the URL parser functionality from the Thunder core here, otherwise this one will assume that the following URL: http:/my.fake.domain.pl/something/https/dial.youtube.com will also pass although I do not thin that that is expected :-) See: https://github.com/rdkcentral/Thunder/blob/master/Source/websocket/URL.h#L28 for URL parsing, I assume that the Type() == SCHEME_HTTPS;, and host portion should ne app.Origin ?

DIALServer/DIALServer.cpp
```diff
@@ -662,16 +662,23 @@ namespace Plugin {
...
+                    safe = (request.Origin.Value().find(app.Origin()) != string::npos);
+            } else if (request.Origin.Value().find("package") == 0) { // Valid format: package: com.youtube.app, package: com.google.ios.youtube
```
pwielders, line 673: Looks like here we also want to make sure that the first work is "package" and the second word is app.origin(). There is a possibility to use a split parser, https://github.com/rdkcentral/Thunder/blob/master/Source/core/TextFragment.h#L547 and use the ':' as the splitter char

pwielders (changes requested):
I think hackers could still insert fake URLs that pass this check... so let it be more strickt. Thunder offeres tools to do so.

DIALServer/DIALServer.cpp
Damian-Danylko: This is examples of origins that are allowed:
https://www.youtube.com, https://dial.youtube.com
package: com.youtube.app, package: com.google.ios.youtube

So you can see that after package it is some combination of app.origin(). where the app.origin() is youtube.com and here we have com.youtube and also other one. So for this reason I simplify the checking procedure. We can think about something more advanced in regards to filtering but it will not be simple app.origin(). Up to you?

DIALServer/DIALServer.cpp
Damian-Danylko: Your url will not be passed because you are using http instead of https :) But I see your point.
In general I think that I have to use Thunder functionality this good approach. But our configuration contains only this portion of host: youtube.com, so there elements like dial. I have to hard-coded for comparison or we will try to adjust configuration to keep more options as allowed origin? I can do my own decision but maybe you would like to have some influence on implementation. Maybe lets have a chat about it?

pwielders: Uhhh I think you are not a real hacker luckily :-) It will pass, since it finds https in the string. although the actual URL is http. Thats what I mean by using URL parsing first and check the scheme, not the path. Your find("https") will find https in the path name, not the scheme. Smae goes for the origin in the path and *not* in the host portion of the URL. Thats why I suggest splitting up the URL in the proper portions Scheme/HOST/[port]/path/query and validate those portions.

Damian-Danylko: I will present you tomorrow YT requirement from Protocol test cases it will be easier to find a proper approach. And I will implement it. I reuse now how it was implemented before so I did not change an approach. Session is scheduled for tomorrow.

DIALServer/DIALServer.cpp
pwielders: So with the URL parser, if it succeeds, just check if the Type() == SCHEME_HTTPS and Host(), last part == app.Origin(), than you definitiely checked the first two, withou the hacker bypassing these checks.
For the package, since the URL parsing will fail (not HTTP scheme) suggest to use the TextSegmentIterator, which will split in the string in two "sections (count() == 2). first must be "package", second part needs discussion I guess.

pwielders (approved):

PR #449, opened 2021-06-28
[METROL-297] Add automatic scanning upon connect/disconnect

lukiwa:
When there is a saved config of the Wifi, there is a chance that one can first scan networks (in which the saved network can not appear for some reason, maybe not in range), and after some time trying to connect to a wifi network that did not appear in searches. That will fail because connect method will not be able to get the BSSID of such a network. The user would need first to scan, and then try to connect to Wifi. 
I resolved this issue by first scanning for networks in the connect method.

### DIFF
--- a/WifiControl/Controller.h
+++ b/WifiControl/Controller.h
@@ -1540,13 +1540,6 @@ namespace WPASupplicant {
         {
             uint32_t result = Core::ERROR_UNKNOWN_KEY;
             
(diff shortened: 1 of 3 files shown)


lukiwa:
Closing, as the better solution seems to be storing BSSID along with the SSID and password in the persistent storage. Then there will be no need to rescan upon every connect request - let higher-level applications decide when to scan.

PR #459, opened 2021-07-30
remove non needed local variable

wiswis15:
small improvement: removed not used local variable


MFransen69:
Hi Wissem as it is a bit more than removing a non needed local var, have you tested this change?

wiswis15:
> Hi Wissem as it is a bit more than removing a non needed local var, have you tested this change?
Yes, in the pi.

MFransen69 (approved):

PR #462, opened 2021-08-03
Development/metrol 333

SKumarMetro:


lukiwa (approved):

ResourceMonitor/ResourceMonitorImplementation.cpp
```diff
@@ -335,11 +336,11 @@ namespace Plugin {
             while (std::getline(csvFile, line)) {
                 allLines.push_back(line);
             }
-            if (allLines.size() > lastMeasurementsHistory) {
+            if (allLines.size()) {
```
MFransen69, line 339: Please always compare with something (we even do that for bools, e.g. == true as a coding guideline for readeability), so here I guess > 0

MFransen69 (changes requested):

PR #468, opened 2021-09-01
Make HDCP level configurable for RPI Platform

sramani-metro:


DisplayInfo/RPI/PlatformImplementation.cpp
```diff
@@ -223,9 +241,38 @@ class DisplayInfoImplementation : public Exchange::IHDRProperties, public Exchan
...
+
+            static bool GetValue(const Core::JSON::Boolean& jsonValue) {
+                return jsonValue.IsSet() ? jsonValue.Value() : false;
+            }
+
+            static int32_t GetValue(const Core::JSON::DecUInt32& jsonValue) {
+                return jsonValue.IsSet() ? jsonValue.Value() : 0;
+            }
+
+            Core::JSON::String hdcpLevel;
```
MFransen69, line 273: Couldn't we better use Core::JSON::EnumType<> ?

DisplayInfo/RPI/PlatformImplementation.cpp
```diff
@@ -350,6 +397,7 @@ class DisplayInfoImplementation : public Exchange::IHDRProperties, public Exchan
     mutable Core::CriticalSection _adminLock;
 
     Core::WorkerPool::JobType< DisplayInfoImplementation& > _activity;
+    Config _config;
```
MFransen69, line 400: Is there actually a reason to store the config for now? We already store the parse result in _value correct?

MFransen69 (changes requested):

DisplayInfo/RPI/PlatformImplementation.cpp
sramani-metro: I thought having a string in the config file would help understand what it means. But yes, We could.

DisplayInfo/RPI/PlatformImplementation.cpp
sramani-metro: You are right. We dont need to store it at this time. I will change this

DisplayInfo/RPI/PlatformImplementation.cpp
MFransen69: The result for the config file itself would be the same. It is the implementation here that is easier to do (e.g. with support for the conversions etc)

sramani-metro: Yes, it is better to use existing enum conversions. It does simplify the implementation by a great deal. Pushed new commit.

MFransen69 (approved):

PR #472, opened 2021-09-07
Weston: Compositor integration

HaseenaSainul:


Compositor/lib/Wayland/Wayland.cpp
```diff
@@ -235,8 +235,8 @@ namespace Plugin {
             : _config()
             , _compositionClients()
             , _clients()
+            , _active(false)
```
pwielders, line 238: I think this _active has no added value. From what I see in the usage, you can just, in the destructor,  do _job.Stop()  and than _job.Wait(STOPPED, Core::inifinite). This will make sure that the thread is "stopped".

pwielders (changes requested):

Compositor/lib/Wayland/Wayland.cpp
HaseenaSainul: thanks @pwielders for the comment, will check

HaseenaSainul: Updated the code as per comments

pwielders (dismissed):

pwielders (approved):

PR #505, opened 2021-11-01
[development/METROL369]: Change the Deinitialize order of out-of-proc…

brainstem123:
…ess if applicable.

- See also WPEPluginAmazon, WPEPluginNetflix and ThunderNanoServicesRDK


WebServer/WebServer.cpp
```diff
@@ -112,18 +112,24 @@ namespace Plugin {
             _memory->Release();
         }
 
-        // Stop processing of the browser:
-        _server->Release();
+        RPC::IRemoteConnection* connection(_service->RemoteConnection(_connectionId));
 
-        if(_connectionId != 0){
-            RPC::IRemoteConnection* connection(_service->RemoteConnection(_connectionId));
+        // Stop processing of the browser:
```
MFransen69, line 117: Know that this line was already there but technically the webserver is not a browser :)

examples/OutOfProcessPlugin/OutOfProcessPlugin.cpp
```diff
@@ -124,11 +124,24 @@ namespace Plugin {
...
 
+        RPC::IRemoteConnection* connection(_service->RemoteConnection(_connectionId));
+
         // Stop processing of the browser:
-        _browser->Release();
+        VARIABLE_IS_NOT_USED uint32_t result = _browser->Release();
 
-        if(_connectionId != 0){
-            ConnectionTermination(_connectionId);
+        // It should have been the last reference we are releasing,
```
MFransen69, line 132: Now we corrected it for the Deinitialize but for the Initialize it is still incorrect (see ConnectionTermination method). Bit similar to the Cobalt situation above. Let's make it completely correct (certainly as people will use this example as a starting point for their OOP plugin

examples/TestController/TestController.cpp
```diff
@@ -114,10 +114,23 @@ namespace Plugin {
 
         _testControllerImp->TearDown();
 
+        RPC::IRemoteConnection* process(_service->RemoteConnection(_connection));
 
-        _testControllerImp->Release();
-        if (_connection != 0) {
-            ProcessTermination(_connection);
+        VARIABLE_IS_NOT_USED uint32_t result = _testControllerImp->Release();
```
MFransen69, line 119: Same remark here as for the OOP plugin

examples/TestUtility/TestUtility.cpp
```diff
@@ -120,10 +120,23 @@ namespace Plugin {
         }
         _memory = nullptr;
 
-        _testUtilityImp->Release();
+        RPC::IRemoteConnection* process(_service->RemoteConnection(_connection));
 
-        if (_connection != 0) {
-            ProcessTermination(_connection);
+        VARIABLE_IS_NOT_USED uint32_t result = _testUtilityImp->Release();
```
MFransen69, line 125: And here same as above

examples/TestUtility/TestUtility.cpp
```diff
@@ -120,10 +120,23 @@ namespace Plugin {
         }
         _memory = nullptr;
 
-        _testUtilityImp->Release();
+        RPC::IRemoteConnection* process(_service->RemoteConnection(_connection));
```
MFransen69, line 123: Although naming is subjective using process here is rather confusing

examples/TestController/TestController.cpp
```diff
@@ -114,10 +114,23 @@ namespace Plugin {
 
         _testControllerImp->TearDown();
 
+        RPC::IRemoteConnection* process(_service->RemoteConnection(_connection));
```
MFransen69, line 117: process might be a bit confusing as name

MFransen69 (changes requested):

Cobalt/Cobalt.cpp
```diff
@@ -82,7 +82,7 @@ static Core::ProxyPoolType<Web::JSONBodyType<Cobalt::Data>> jsonBodyDataFactory(
                 Exchange::JApplication::Register(*this, _application);
             } else {
                 stateControl->Release();
-                _cobalt->Release();
+                CobaltTermination();
                 _cobalt = nullptr;
```
MFransen69, line 86: Let's also move _cobalt = nullptr; into CobaltTermination, that is what you would expect and then it cannot be forgotten after the CobaltTermination call

Cobalt/Cobalt.cpp
```diff
@@ -92,7 +92,7 @@ static Core::ProxyPoolType<Web::JSONBodyType<Cobalt::Data>> jsonBodyDataFactory(
         message = _T("Cobalt could not be instantiated.");
         _service->Unregister(&_notification);
         _service = nullptr;
-        ConnectionTermination(_connectionId);
+        CobaltTermination();
```
MFransen69, line 95: Here we should not call it anymore. Now Thunder will crash when Cobalt was stopped in the code above as it will do a Release call on the nullptr

examples/OutOfProcessPlugin/OutOfProcessPlugin.cpp
```diff
@@ -63,7 +63,7 @@ namespace Plugin {
             _service->Unregister(static_cast<PluginHost::IPlugin::INotification*>(_notification));
             _service = nullptr;
 
-            ConnectionTermination(_connectionId);
+            OutOfProcessTermination();
```
MFransen69, line 66: Let's not call it here, it will crash Thunder, it should not be needed as the Root call failed

examples/OutOfProcessPlugin/OutOfProcessPlugin.cpp
```diff
@@ -344,10 +326,24 @@ namespace Plugin {
...
     }
 
-    void OutOfProcessPlugin::ConnectionTermination(uint32_t connectionId)
+    void OutOfProcessPlugin::OutOfProcessTermination()
     {
-        RPC::IRemoteConnection* connection(_service->RemoteConnection(connectionId));
+        RPC::IRemoteConnection* connection(_service->RemoteConnection(_connectionId));
+
+        // Stop processing of the browser:
+        VARIABLE_IS_NOT_USED uint32_t result = _browser->Release();
```
MFransen69, line 334: let's now also set _browser to nullptr here

examples/TestController/TestController.cpp
```diff
@@ -88,13 +88,14 @@ namespace Plugin {
...
                 connection->Release();
+                OutOfProcessTermination();
                 _testControllerImp->Setup();
             } else {
                 _memory = nullptr;
                 TRACE(Trace::Warning, (_T("Colud not create MemoryObserver in TestController")));
             }
         } else {
-            ProcessTermination(_connection);
+            OutOfProcessTermination();
```
MFransen69, line 98: In practice service cannot be a nullptr so it will only get here if the Rool call failed and as above then it will crash as it will try to call Release on the nullptr

examples/TestController/TestController.cpp
```diff
@@ -88,13 +88,14 @@ namespace Plugin {
                 ASSERT(_memory != nullptr);
 
                 connection->Release();
+                OutOfProcessTermination();
```
MFransen69, line 91: I'm not sure why you add the call here? The next line still uses it (and will now crash :) )

examples/TestController/TestController.cpp
```diff
@@ -185,12 +169,25 @@ namespace Plugin {
...
-    void TestController::ProcessTermination(uint32_t connection)
+    void TestController::OutOfProcessTermination()
     {
-        RPC::IRemoteConnection* process(_service->RemoteConnection(connection));
-        if (process != nullptr) {
-            process->Terminate();
-            process->Release();
+        RPC::IRemoteConnection* connection(_service->RemoteConnection(_connectionId));
+
+        VARIABLE_IS_NOT_USED uint32_t result = _testControllerImp->Release();
```
MFransen69, line 176: Let's also set _testControllerImp to nullptr after this

examples/TestController/TestController.cpp
```diff
@@ -185,12 +169,25 @@ namespace Plugin {
         return result;
     }
 
-    void TestController::ProcessTermination(uint32_t connection)
+    void TestController::OutOfProcessTermination()
```
MFransen69, line 172: As it might be confusing better name the method differently (of course I also would have copied it from the previous one :) )

examples/TestUtility/TestUtility.cpp
```diff
@@ -190,21 +174,31 @@ namespace Plugin {
         return result;
     }
 
-    void TestUtility::ProcessTermination(uint32_t _connection)
+    void TestUtility::OutOfProcessTermination()
```
MFransen69, line 177: as above, this name might be confusing

examples/TestUtility/TestUtility.cpp
```diff
@@ -190,21 +174,31 @@ namespace Plugin {
...
     {
-        RPC::IRemoteConnection* process(_service->RemoteConnection(_connection));
+        RPC::IRemoteConnection* connection(_service->RemoteConnection(_connectionId));
 
+        VARIABLE_IS_NOT_USED uint32_t result = _testControllerImp->Release();
```
MFransen69, line 181: Don't think this even compiles :) it should be _testUtilityImp if I'm right. And let's also set it to nullptr after this

examples/TestUtility/TestUtility.cpp
```diff
@@ -73,29 +73,30 @@ namespace Plugin {
...
                 remoteConnection->Release();
+                OutOfProcessTermination();
             } else {
                 _memory = nullptr;
                 TRACE(Trace::Warning, (_T("Colud not create MemoryObserver in TestUtility")));
             }
 
         } else {
-            ProcessTermination(_connection);
+            OutOfProcessTermination();
```
MFransen69, line 99: And calling it here will crash Thunder

examples/TestUtility/TestUtility.cpp
```diff
@@ -73,29 +73,30 @@ namespace Plugin {
...
+        _testUtilityImp = _service->Root<Exchange::ITestUtility>(_connectionId, ImplWaitTime, _T("TestUtilityImp"));
 
         if (_testUtilityImp != nullptr) {
-            RPC::IRemoteConnection* remoteConnection = _service->RemoteConnection(_connection);
+            RPC::IRemoteConnection* remoteConnection = _service->RemoteConnection(_connectionId);
 
             if (remoteConnection) {
                 _memory = WPEFramework::TestUtility::MemoryObserver(remoteConnection);
                 remoteConnection->Release();
+                OutOfProcessTermination();
```
MFransen69, line 92: Do not understand why it was added here, makes the plugin no longer work :)

MFransen69 (changes requested):

PR #506, opened 2021-11-09
Enable setting of RemoteControl config values

Michal-Bugno:
Enable setting of repeat interval and repeat start in RemoteControl, so it is not hard-coded


artur-gebicz-ml (approved):

artur-gebicz-ml:
@pwielders could you please make sure, that this PR lands on the R3 branch some day?

PR #509, opened 2021-11-12
HashVerification moved to Download thread, instead doing it from the webtransfer thread

HaseenaSainul:


HaseenaSainul:
Closing, since this changes is not required the expected changes is only thunder alone

PR #538, opened 2022-01-27
DIALServer: Read DeviceId from the subsystem

HaseenaSainul:


DIALServer/DIALServer.h
```diff
@@ -1273,6 +1273,30 @@ namespace Plugin {
             return (_deprecatedAPI);
         }
 
+        string DeviceId() const
+        {
+            string deviceId;
+            PluginHost::ISubSystem* subSystem = _service->SubSystems();
```
pwielders, line 1279: Be carefull, this is a leak. If you get the subSystem, it is an addref'ed interface make sure you release it!

HaseenaSainul: done

pwielders: Sorry, not completly, it is still leaking if the subsystem is retrieved (not a nullptr), it is addref'ed but if the SubSystem->IsActive() == false, it never gets released and thus it is leaking.. Suggest t remove the IsActive check from the equation and it is good to go...

pwielders (changes requested):

pwielders (changes requested):
It is still leaking if the system is not active, suggest to remove the IsActive call or split the if..

DIALServer/DIALServer.h
HaseenaSainul: done

HaseenaSainul:
IsActive check is removed

pwielders (approved):

PR #561, opened 2022-03-15
[WebServer] IStateControl -> IConfigure

MFransen69:


pwielders (approved):

MFransen69:
In the end the IStateControl was there for another reason than just the Configure call. See https://jira.rdkcentral.com/jira/browse/METROL-489

PR #567, opened 2022-03-23
Dictionary: fix issue with save data to disk

HaseenaSainul:


Dictionary/Dictionary.cpp
```diff
@@ -132,10 +132,12 @@ namespace Plugin {
 
         Core::File dictionaryFile(service->PersistentPath() + _config.Storage.Value());
 
-        if (dictionaryFile.Open(true) == true) {
+        if (dictionaryFile.Create() == true) {
             NameSpace dictionary;
             CreateExternalDictionary(EMPTY_STRING, dictionary);
-            dictionary.IElement::ToFile(dictionaryFile);
+            if (dictionary.IElement::ToFile(dictionaryFile) == false) {
+                TRACE(Trace::Warning, (_T("Error occured while trying to save dictionary data to file!")));
```
pwielders, line 139: Let;s make this a SYSLOG, Shutdown category as this is quite crucial to report.

pwielders (changes requested):

Dictionary/Dictionary.cpp
HaseenaSainul: done

pwielders (approved):

PR #570, opened 2022-03-29
Test App: StoreTest added to test Performance of PersistentStore and Dictionary plugin via JSONRPC, ReST and COMRPC channels

HaseenaSainul:


examples/StoreTest/StoreTest.cpp
```diff
@@ -0,0 +1,558 @@
...
+
+private:
+    void Operational(const bool upAndRunning)
+    {
+        printf("Operational state of PersistentStore: %s\n", upAndRunning ? _T("true") : _T("false"));
+    }
+};
+
+namespace JSONRPC {
+class PersistentStore : public Thunder::JSONRPC::SmartLinkType<Thunder::Core::JSON::IElement> {
```
MFransen69, line 174: If we are going to put this in examples people will consider this code that reflects correctly how things should be done and will copy this code as a starting point. So either we will make it so and use JSONRPC::LinkType (see the jsonrpocclient example) and certainly not JsonObject (which is not allowed in ML code) or we just put this code next to our unittest and other test code. (perhaps the latter is a better approach all together). I think it is more of a test then an example, it is also called StoreTest :) so I guess better to out it with the tests

examples/StoreTest/StoreTest.cpp
```diff
@@ -0,0 +1,558 @@
...
+template <typename STORE>
+class StorePerformance {
+
+public:
+    StorePerformance(uint32_t waitTime, const Thunder::Core::NodeId& nodeId, const string& callSign)
+        : _store(waitTime, nodeId, callSign)
+        , _callSign(callSign)
+    {
+    }
+    virtual ~StorePerformance() = default;
```
MFransen69, line 395: why is this virtual? Perhaps I missed it but there does not seem to be any virtual method

examples/StoreTest/StoreTest.cpp
```diff
@@ -0,0 +1,558 @@
...
+private:
+    using BaseClass = Thunder::RPC::SmartInterfaceType<Thunder::Exchange::IDictionary >;
+public:
+    Dictionary(const uint32_t waitTime, const Thunder::Core::NodeId& node, const string& callsign)
+        : BaseClass()
+    {
+        BaseClass::Open(waitTime, node, callsign);
+        BaseClass::Aquire<Thunder::Exchange::IDictionary>(waitTime, node, callsign);
+    }
+    ~Dictionary()
```
MFransen69, line 86: nitpicking I guess, but it should be overriden, please also check below (if we put it in examples it should be rather correct)

examples/StoreTest/StoreTest.cpp
```diff
@@ -0,0 +1,558 @@
...
+            string ns = "/" + nameSpace;
+            result = impl->Set(ns, key, value);
+            impl->Release();
+        }
+
+        return (result);
+    }
+
+private:
+    void Operational(const bool upAndRunning)
```
MFransen69, line 120: think this is an override as well (also check below)

MFransen69 (changes requested):

examples/StoreTest/StoreTest.cpp
HaseenaSainul: Ok, Will move this to unit tests folder

examples/StoreTest/StoreTest.cpp
HaseenaSainul: will correct

examples/StoreTest/StoreTest.cpp
HaseenaSainul: will check

examples/StoreTest/StoreTest.cpp
HaseenaSainul: will check

examples/StoreTest/StoreTest.cpp
HaseenaSainul: I changed JsonObject to JSON::Container based class based on Parameter and Response requirement.
I used JSONRPC::SmartLinkType instead JSONRPC::LinkType to get the plugin activated/deactivated status to return proper states from IsOperational() method.

HaseenaSainul: Will move this to Thunder as next step :)

HaseenaSainul: Moved StoreTest to tests

MFransen69 (approved):

PR #574, opened 2022-04-26
Bug fix and feature enhancement for Power plugin

susanmary007:
This PR contains 4 commits:

f18459473340879c9660003c8e359b72d9c5c266 allows to send power state change notification before and after power state change.

f18459473340879c9660003c8e359b72d9c5c266 - ad98733c498319bc2be713c4aaed54514664b37b fix some issues observed


Power/Power.cpp
```diff
@@ -263,6 +264,10 @@ namespace Plugin {
     }
     void Power::StateChange(PluginHost::IShell* plugin)
     {
+        if (!_controlClients) {
```
pwielders, line 267: I would consider this to be a precondition. Suggest to ASSERT, in debug this will fire. In release it will not. But if this ASSERt fires, it means that the Register happened although the initialize checks this value.

pwielders, line 267: This is against the coding guidelines (only 1 return allowed, for maintainability) But I also think that this is a pre-requisite and thus could be ASSERTED. ASSERT(_controlClients) cause this method has a direct relation to the Register/Unregister. In the Register you check if the _ControlClients is true and only than register, so Programitcally this could never happen unless there is a prgramming mistake, which can be handled with an ASSERT.

pwielders (changes requested):
Minor topic, Code rule violation which can probably be avoied by issueing an ASSERT.

Power/Power.cpp
susanmary007: Agree.. Will make the changes

susanmary007:
Created v2(https://github.com/rdkcentral/ThunderNanoServices/pull/580)

PR #576, opened 2022-04-26
Implementing Suspend and Resume for webserver

Karthick-Somasundaresan:


WebServer/WebServerImplementation.cpp
```diff
@@ -668,35 +685,71 @@ POP_WARNING()
...
-                result = _channelServer.Open(2000);
-            }
-
+            uint32_t result = _channelServer.Configure(service->DataPath(), config);
             return (result);
         }
         PluginHost::IStateControl::state State() const override
         {
-            return (PluginHost::IStateControl::RESUMED);
+            PluginHost::IStateControl::state state;
```
pwielders, line 710: Minor, no need to change but for these short construction it is oke to use the Core::SafeSync<Core::CriticalSection> lock(_adminLock);

WebServer/WebServerImplementation.cpp
```diff
@@ -668,35 +685,71 @@ POP_WARNING()
...
             : _channelServer()
+            , _adminLock()
             , _observers()
+            , _state(PluginHost::IStateControl::UNINITIALIZED)
         {
         }
 
-        ~WebServerImplementation() override = default;
+        ~WebServerImplementation() override {
+            for (auto & observer: _observers){
```
pwielders, line 695: There is no need to do this. Better would be to ASSERT to indicate a potential memory leak and then do run through the observeers, but only Release() the interface(), although that will probably nolonger work :-)

pwielders (changes requested):
There is a race condition in the code that requires some attention.

MFransen69:
We'll pick this up in the future again, lower priority for now

PR #578, opened 2022-05-09
Enable AVS Smart Screen Define and add config for DB path

HaseenaSainul:


AVS/Integration/CMakeLists.txt
```diff
@@ -15,12 +15,21 @@
...
+
 # platform independent files
 install(DIRECTORY
     DESTINATION "${PLUGIN_AVS_DATA_PATH}/${PLUGIN_AVS_NAME}"
     COMPONENT ${MODULE_NAME})
 
 # platform dependent files
+file(READ ${PLUGIN_AVS_PLATFORM}/AlexaClientSDKConfig.json.in MOD_ALEXA_CLIENT_SDK_CONFIG)
+    string(
+           REGEX REPLACE "AVS_DB_PATH" "${PLUGIN_AVS_DB_PATH}"
```
artur-gebicz-ml, line 28: Could you please wrap it with some sort of symbols? It will prevent replacing AVS_DB_PATH when not used as a token and will clearly be visible which part of the config will be repalaced. Something like <AVS_DB_PATH> or ##AVS_DB_PATH## etc.

AVS/Integration/CMakeLists.txt
```diff
@@ -15,12 +15,21 @@
...
 
+set(PLUGIN_AVS_DB_PATH "/root/AVS/db" CACHE STRING "AVS database directory location")
+
 # platform independent files
 install(DIRECTORY
     DESTINATION "${PLUGIN_AVS_DATA_PATH}/${PLUGIN_AVS_NAME}"
     COMPONENT ${MODULE_NAME})
 
 # platform dependent files
+file(READ ${PLUGIN_AVS_PLATFORM}/AlexaClientSDKConfig.json.in MOD_ALEXA_CLIENT_SDK_CONFIG)
```
artur-gebicz-ml, line 26: Would it be possible to use configure_file instead? https://cmake.org/cmake/help/latest/command/configure_file.html

artur-gebicz-ml (changes requested):

AVS/Integration/CMakeLists.txt
HaseenaSainul: updated

AVS/Integration/CMakeLists.txt
HaseenaSainul: updated

artur-gebicz-ml (approved):

PR #583, opened 2022-05-17
Development/cec control plugin

bramoosterhuis:


mhughesacn:
Compliance note if this comes up again:
There were two code matches in this PR.  Both were removed by renaming variables: CECTypes.h for CEC User Control codes, and AmlogicAdapter.cpp for CEC iotcl macros.  The code match was to GPLv2 code.  In both cases, the matches could be considered boilerplate. The user control codes were straight from the HDMI spec.  The macros are used in interfacing with the linux kernel.  Renaming is perfectly reasonable.

pwielders (approved):

pwielders:
@mhughesacn thanks for the feedback! We were not sure..

PR #587, opened 2022-05-20
Development/rust bridge

pwielders:


mhughesacn:
Hi Pierre: Please use Metrological headers.  Is there any reason they aren't on the two js files? - Thanks, Martin

pwielders:
Hi Martin, its due to the fact that they where contributed by someone from Comcast. They are example files. Let me check with Comcast if I should copyright them with Metrological or with Comcast. or Would Comcast be a problem ?

pwielders:
@mhughesacn now looked at the git report, it is the wrong disclaimer in our code ;-) No problem, I will correct it.

pwielders:
@mhughesacn is it a problem that the example code does not have a license header ?

mhughesacn:
Hi @pwielders RDK policy is that all code files have headers.  It isn't a blocker though and if you have a justification, please tell me.
Also, for most components, using Comcast header on Comcast-created files is fine.  But if they are changed, use a Metro header on top (and maybe truncate the Comcast header to a two liner "Copyright Year Comcast.../Licensed under the Apache License").

MFransen69 (approved):

PR #593, opened 2022-06-08
Cobalt platform services support to send / recv messages from JS.

neeraj-deshpande:


MFransen69:
We agreed to pause this for now. We will add this once there is a real use case.

PR #604, opened 2022-07-26
[BT] Make sure we have a verbose trace that reports byte streams..

pwielders:


TimeSync/TimeSync.config
```diff
@@ -1,4 +1,5 @@
 set(autostart ${PLUGIN_TIMESYNC_AUTOSTART})
+set(preconditions Internet)
```
HaseenaSainul, line 2: Currently we removed the precondition settings from the config to code. So do we need to keep still it in the config ?

pwielders: Yes as currently the only client is an NTP server and typicall the address we use here are internaet addresses. So the default that is part of the timesync (I assume is network only) whihc is good since the NTP server could be on the local net.
Will do a presentation on this during the netx WorlWide Team meeting and will use this as an example.

HaseenaSainul: Ok, but can we move this preconditions::Internet to TimeSync.cpp ? instead TimeSync.config ?

pwielders: No, because if the NTPServer is in the local network we do not need the internet to correct the time. SO the metadata should only list network connectivity used and the config (as we are adding typically som internet ntp servers) should than list the fact that we also need internet connectoivity.

HaseenaSainul: Ok, then in this case, do we need to keep preconditions::Internet  in  TimeSync.config? can we move to  TimeSync.cpp, since we are setting this as absolute precondition ?

pwielders: In the plugin itself (TimeSync.cpp) the metdata section, it has to be internet. Depending on the servers we fill in to synchronize the time with, we might need to add the precondition internet in the consfig. If, like the config is today, we fill in:
map_append(${configuration} sources ntp://0.pool.ntp.org)
map_append(${configuration} sources ntp://1.pool.ntp.org)
map_append(${configuration} sources ntp://2.pool.ntp.org)
map_append(${configuration} sources ntp://3.pool.ntp.org)

we know that these are services on the internet. So internet connectivity is required. Since the plugin itself is only expecting NETWORK (not INTERNET), we should add INTERNET as a preconditon as well in the config.

 Now suppose we change the config to:
 map_append(${configuration} sources ntp://mygateway.net)

it will use the NTP server available in your home router, which is on the local network. In that case you do not need INTERNET access, you only need NETWORK access. So no need to build the plugin again (as it has in the METADATA only NETWORK) and you should only remove the precondition INTERNET from the config.

So with the given default servers, the precondition is correct, and the metadata section in the TIMESYNC should be NETWORK.

HaseenaSainul: Thanks for the clarification

HaseenaSainul (approved):

PR #623, opened 2022-08-24
ResourceMonitor: set CSVFile Path to read values from CompileMemoryCsv

HaseenaSainul:


lukiwa (approved):

ResourceMonitor/ResourceMonitorImplementation.cpp
```diff
@@ -313,6 +314,10 @@ namespace Plugin {
             Config config;
             config.FromString(service->ConfigLine());
 
+            if (config.Path.IsSet()) {
```
pwielders, line 317: Can we, by default, prefix the given name ( or the default name) with Service->VolatilePath. If the system is configured to have the temporary filesystem configured somewhere else than /temp, it still works as expected :-)

pwielders (changes requested):

ResourceMonitor/ResourceMonitorImplementation.cpp
HaseenaSainul: Sure, will update :)

pwielders (approved):

PR #646, opened 2022-09-20
ConfigUpdateExample: plugin added to show how to update plugin config…

HaseenaSainul:
… systemrootpath and persist


examples/ConfigUpdateExample/ConfigUpdateExample.cpp
```diff
@@ -0,0 +1,84 @@
...
+	if (config.Plugins.IsSet() == true) {
+            Core::JSON::ArrayType<Config::Plugin>::Iterator index(config.Plugins.Elements());
+	    while (index.Next() == true) {
+                 const string& callsign(index.Current().Callsign.Value());
+                 PluginHost::IShell* shell = service->QueryInterfaceByCallsign<PluginHost::IShell>(callsign);
+		 if (shell != nullptr) {
+		     printf("Current SystemRootPath[%s] = %s\n", callsign.c_str(), shell->SystemRootPath().c_str());
+		     if (index.Current().SystemRootPath.IsSet() == true) {
+                         shell->SystemRootPath(index.Current().SystemRootPath.Value());
+    		         printf("New SystemRootPath[%s] = %s\n", callsign.c_str(), shell->SystemRootPath().c_str());
```
pwielders, line 58: Can we turn the printf's into TRACE statements.

examples/ConfigUpdateExample/ConfigUpdateExample.cpp
```diff
@@ -0,0 +1,84 @@
...
+                         PluginHost::IController* controller = service->QueryInterfaceByCallsign<PluginHost::IController>(EMPTY_STRING);
+                         if (controller != nullptr) {
+                             controller->Persist();
```
pwielders, line 61: Persist will always store all plugins in one file, so it should be sufficient to call this outside the while loop.

HaseenaSainul: done

examples/ConfigUpdateExample/ConfigUpdateExample.cpp
HaseenaSainul: done

pwielders (approved):

PR #661, opened 2022-09-30
RemoteControl: Fix potential deadlock

jfifranca:
Fixing potential deadlock when registering/unregistering for events, where if notification client is already listed/non-existing it would leave the Core::CriticalSection _eventLock in permanent lock state, thus locking all event notifications.


pwielders (approved):

pradeeptakdas:
https://jira.rdkcentral.com/jira/browse/RDKDEV-355

PR #673, opened 2022-10-28
[SYNC] Refactor of the backoffice plugin.

pwielders:


BackOffice/BackOffice.config
```diff
@@ -1,4 +1,5 @@
 set(autostart ${PLUGIN_BACKOFFICE_AUTOSTART})
+set(preconditions Platform Internet)
```
HaseenaSainul, line 2: Why added it back, since we are handling this from MetaData

BackOffice/BackOffice.config
```diff
@@ -25,9 +26,5 @@ map()
 end()
 ans(configuration)
 
-if(PLUGIN_BACKOFFICE_CALLSIGN_MAPPING)
```
HaseenaSainul, line 28: Please ensure the callsign mapping is not empty, otherwise it will create a wrong json line like, 
like
"callsign_mapping":,

BackOffice/BackOffice.config
```diff
@@ -25,9 +26,5 @@ map()
...
 map_append(${configuration} callsign_mapping ___array___ ${PLUGIN_BACKOFFICE_CALLSIGN_MAPPING})
-endif()
-if(PLUGIN_BACKOFFICE_STATE_MAPPING)
```
HaseenaSainul, line 31: Above comment is applicable here as well

BackOffice/RequestSender.h
```diff
@@ -1,149 +0,0 @@
...
-                        auto entry = _parent._queue.back();
-                        _parent._queue.pop_back();
-
-                        Trace::Information(_T("State change to open, attempting to send a request"));
-                        auto request = Core::ProxyType<WPEFramework::Web::Request>::Create();
-                        request->Verb = Web::Request::HTTP_GET;
-                        request->Query = Core::Format("%sevent=%s&id=%s", _parent._queryParameters.c_str(), entry.first.c_str(), entry.second.c_str());
-                        request->Host = _parent._hostAddress;
-                        request->Accept = _T("*/*");
-                        request->UserAgent = _parent._userAgent;
```
HaseenaSainul, line 69: We don't need userAgent ?

BackOffice/BackOffice.h
```diff
@@ -3,308 +3,383 @@
...
-            void UnregisterOperationalNotifications(PluginHost::IShell* service)
-            {
-                ASSERT(service != nullptr);
-                _callback = nullptr;
-                service->Unregister(&_operationalObserver);
+                Core::JSON::ArrayType<Core::JSON::String>::ConstIterator index = config.CallsignMapping.Elements();
+
+                // Check out for which Callsigns we need to report the lifetime stuff.
+                while (index.Next() == true) {
+                    // TODO: To make it more clear it is anothername for the callsign, suggest to use an euqla sign
```
HaseenaSainul, line 264: euqla = equal ?

BackOffice/BackOffice.config
pwielders: Good point! Let me correct it

BackOffice/BackOffice.config
pwielders: Corrected!

BackOffice/BackOffice.config
pwielders: Corrected!

BackOffice/RequestSender.h
pwielders: This is a very good catch. The BackOffice plugin has been lingering around for a long time and during the testing of the other one it was observed that the UserAgent and Accept where mandatory for the Metrological BO servers to pick these up and thus they where added during the last phase but never endede up on this. So Very good catch ! Thanks !

HaseenaSainul (approved):

PR #693, opened 2023-01-03
_service call usage changes

HaseenaSainul:


TimeSync/TimeSync.cpp
```diff
@@ -76,26 +76,25 @@ POP_WARNING()
...
         // On success return empty, to indicate there is no error text.
         return _T("");
     }
 
-    /* virtual */ void TimeSync::Deinitialize(PluginHost::IShell* service VARIABLE_IS_NOT_USED)
+    /* virtual */ void TimeSync::Deinitialize(PluginHost::IShell* service)
     {
-        if (_service != nullptr) {
-            ASSERT(_service == service);
+        if (service != nullptr) {
```
pwielders, line 89: why would you check if (serrvice != nullptr) ??? can imagine you would ASSERT on int by stating ASSERT(service != nullptr), but bottomline you can always execute the code as it does not require the service parameter..

TimeSync/TimeSync.cpp
```diff
@@ -204,16 +203,12 @@ POP_WARNING()
...
     void TimeSync::EnsureSubsystemIsActive()
     {
-        ASSERT(_service != nullptr);
-        PluginHost::ISubSystem* subSystem = _service->SubSystems();
         ASSERT(subSystem != nullptr);
 
-        if (subSystem != nullptr) {
-            if (subSystem->IsActive(PluginHost::ISubSystem::TIME) == false) {
-                subSystem->Set(PluginHost::ISubSystem::TIME, _client);
+        if (_subSystem != nullptr) {
```
pwielders, line 208: Minor one, but from code perspective, you could ASSERT here using ASSERT(_subsystem != nullptr) and than in the Initialize return an Error if the parameter is a nullptr. Why would you initialize a TimeSync if it can not set the Subsystem ?

pwielders (commented):
Minor issues, but maybe good to think about ...

TimeSync/TimeSync.cpp
HaseenaSainul: done

TimeSync/TimeSync.cpp
HaseenaSainul: done

pwielders (approved):

PR #734, opened 2023-06-14
[COMRPC] Make sure the AnnounceMessage is used correctly.

pwielders:


examples/COMRPCClient/COMRPCClient.cpp
```diff
@@ -91,6 +92,34 @@ MODULE_NAME_DECLARATION(BUILD_REFERENCE);
 
 namespace Thunder = WPEFramework;
 
+class Sink : public Thunder::Exchange::IDictionary::INotification {
+public:
+    Sink() = delete;
+    Sink(Sink&&) = delete;
+    Sink(const Sink&) = delete;
+    Sink& operator= (const Sink&) = delete;
```
MFransen69, line 100: see other review,l although technically not needed we discussed also deleting the move operator

MFransen69 (approved):

PR #752, opened 2023-09-19
Introduce VaultProvisioning

sebaszm:
Based on config VaultProvisioning will create a encrypted blob, that will be picked up by Svalbard on startup and loaded into the appropriate vault. Currently only provisioning of randomly generated key is supported.


mhughesacn:
Hi Sebastian: Blackduck is complaining about RDK Apache headers instead of Metrological.  Is there a reason for these?
Thanks.

sebaszm:
> Hi Sebastian: Blackduck is complaining about RDK Apache headers instead of Metrological. Is there a reason for these? Thanks.

Hi Martin, the reason is clumsy copy & paste. Fixed, thanks!

mhughesacn:
Many thanks Sebastian.

pwielders (approved):

PR #761, opened 2024-01-02
[NAMING] Update the Service and Sink to their respective template def…

pwielders:
…initions ServiceType and SinkType.

Depending on: https://github.com/rdkcentral/Thunder/pull/1478


HaseenaSainul (dismissed):

pwielders:
We need: https://github.com/rdkcentral/ThunderNanoServices/pull/760 to succeed before we can succeed!

HaseenaSainul (approved):

PR #772, opened 2024-01-18
Remove NEGATIVE settings on PluginHost::ISubSystem from the plugins deinit sequence

HaseenaSainul:


pwielders (approved):
@HaseenaSainul I see this is a good fix as currently the Framework "turns off" the preconditions and it has to since this has to be done *before* we really deactivate the plugin. 
However this brings in some a-symmetrical behaviour which I think we should document. The setting of the subsystem is delegated to the Plugin Initialize (or operation) and the deactivation is not. 
Could you raise a JIRA ticket so we will document this and reference than to this PR  and JIRA ticket associated with this PR as well ?
Thanks!

HaseenaSainul:
> @HaseenaSainul I see this is a good fix as currently the Framework "turns off" the preconditions and it has to since this has to be done _before_ we really deactivate the plugin. However this brings in some a-symmetrical behaviour which I think we should document. The setting of the subsystem is delegated to the Plugin Initialize (or operation) and the deactivation is not. Could you raise a JIRA ticket so we will document this and reference than to this PR and JIRA ticket associated with this PR as well ? Thanks!

JIRA ticket is created : https://jira.rdkcentral.com/jira/browse/METROL-927

PR #786, opened 2024-04-03
Register/Unregister: ensure sink exists in the list or not before registering and unregistering

HaseenaSainul:


NetworkControl/NetworkControlImplementation.cpp
```diff
@@ -700,23 +700,32 @@ namespace Plugin
         uint32_t Register(Exchange::INetworkControl::INotification* notification) override
         {
             ASSERT(notification);
-            _adminLock.Lock();
-            notification->AddRef();
-            _notifications.push_back(notification);
-            _adminLock.Unlock();
+
+            if (notification) {
```
pwielders, line 704: Lets apply the same pattern on Register/Unregister everywhere. We ASSERT in debug on notification == nullptr, but after that we do insert it. This one is deviating from it. 
1) Or all registers we check for ((<notificaition> != nullptr) && (<iterator> == <list>.end())) { Register() }
2) Or all registers only on (<iterator> == <list>.end())
To be honoust as we check on already "registered", I think it is usefull to *also* check for nullptr, but that is currently more work so can also live with option 2 ;-) But lets not mix :-)

NetworkControl/NetworkControlImplementation.cpp
```diff
@@ -700,23 +700,32 @@ namespace Plugin
...
         uint32_t Unregister(Exchange::INetworkControl::INotification* notification) override
         {
             ASSERT(notification);
-            _adminLock.Lock();
-            auto item = std::find(_notifications.begin(), _notifications.end(), notification);
-            ASSERT(item != _notifications.end());
-            _notifications.erase(item);
-            (*item)->Release();
-            _adminLock.Unlock();
+            if (notification) {
```
pwielders, line 721: See above, this is deviating from what all the others did....

pwielders (changes requested):

NetworkControl/NetworkControlImplementation.cpp
HaseenaSainul: updated

NetworkControl/NetworkControlImplementation.cpp
HaseenaSainul: done

pwielders (approved):

PR #818, opened 2024-07-24
Compositor: avoid additional addref to avoid dangling proxies during the exit of applications

HaseenaSainul:


HaseenaSainul:
Closing, since this is not valid after adding Dangling override methods

PR #820, opened 2024-08-01
example compile fixes + cleanup

HaseenaSainul:


pwielders (approved):
So thats "Good Busy!" Looks Good !

PR #834, opened 2024-10-01
[WebServer] Fixing an invalid iterator

VeithMetro:
The iterator was being invalidated after each [erase](https://github.com/rdkcentral/ThunderNanoServices/blob/master/WebServer/WebServerImplementation.cpp#L810) in `Unregister()`, which caused an undefined behavior like calling the method more times than expected and led to an ASSERT during the framework shutdown

### DIFF
--- a/WebServer/WebServerImplementation.cpp
+++ b/WebServer/WebServerImplementation.cpp
@@ -717,9 +717,10 @@ POP_WARNING()
         {
         }
 
-        ~WebServerImplementation() override {
-            for (auto & observer: _observers){
-                Unregister(observer);
+        ~WebServerImplementation() override
+        {
+            while (!_observers.empty()) {
+                Unregister(_observers.front());
             }
         }
 


sebaszm (approved):

sebaszm (approved):
Still good change

WebServer/WebServerImplementation.cpp
```diff
@@ -796,6 +798,7 @@ POP_WARNING()
             // Only subscribe an interface once.
             std::list<PluginHost::IStateControl::INotification*>::iterator index(std::find(_observers.begin(), _observers.end(), notification));
 
+            _adminLock.Lock();
```
MFransen69, line 801: Eh am I missing something perhaps, shouldn't we take the lock before line 799 :) there we also already use _observers ;)

MFransen69 (changes requested):

WebServer/WebServerImplementation.cpp
VeithMetro: Something must've went wrong during the merging of master branch, as it was definitely in the correct spot before.. fixed!

MFransen69 (approved):

PR #851, opened 2024-11-19
Development/202410 compositor

bramoosterhuis:
This will add:

- Double buffered frame buffers with the swap timed on the V Sync.
- Adds the Atomic DRM API
- Fixed opacity setting.


pwielders (approved):
Lets merge it. It is only affecting the MESA compositor

PR #891, opened 2025-03-28
Development/reuse output framebuffer

bramoosterhuis:


pwielders (changes requested):
I think the IGPU.h (local interface fie) can be dropped now a well..

bramoosterhuis:
> I think the IGPU.h (local interface fie) can be dropped now a well..

But that is refactor specific to the DRM back-end, these changes are about the render engine not allocating a FrameBuffer object on every frame. So the proposed change feels unrelated to that. 

Still I think it's a improvement we should do, but not here.

pwielders (approved):

PR #893, opened 2025-04-08
Development/robustness fixes

bramoosterhuis:


Compositor/lib/Mesa/Compositor.cpp
```diff
@@ -279,7 +288,44 @@ namespace Plugin {
              */
             void Request() override
             {
-                _parent.Render();
+                _renderLock.lock();
```
pwielders, line 291: What is the purpose of the renderlock. I guess f you render, you Acquire the plane which you would like to render (or at least the purpose of this lock is to let the other side know, this is a cross process lock). So the buffer from which to render is locked. I think it is safe to drop this lock and leave it up to the buffer lock..

Compositor/lib/Mesa/Compositor.cpp
```diff
@@ -279,7 +288,44 @@ namespace Plugin {
...
+
+                State expected(State::IDLE);
+
+                std::lock_guard<std::mutex> lock(_lock);
+                if (_state.compare_exchange_strong(expected, State::RENDERING) == true) {
+                    _parent.Render(); // Trigger rendering operations
```
pwielders, line 297: Oke the Renderer is now a-synchronous. Than I guess it is even more important to "use" the buffer lock since now the actual render might take place at a much later time, when the client might be using this surface as well again.
Would it not be simler to synchronously execute the render at the moment we see it is required? It makes the flow simpler, no a-synchronous behavior. 
You can also say, we are only going to Render after a V-Sync (as you do now) but than you have to make sure that the client is aware (and thus use the cross process lock, on the Buffer and not a local lock _renderLock)
A crucial question here to answer is "why" wait with the Render till you saw a v-sync. The output is double buffered, in this process the double buffered switch is done, so for this you could use a mutex (if needed). However if it is all running on the resource monitor thread it is serialized anayway so no lock needed at all.. If you run your own thread yes use a mutex.

Compositor/lib/Mesa/Compositor.cpp
```diff
@@ -907,23 +918,36 @@ namespace Plugin {
...
                 uint32_t result(Core::ERROR_NONE);
 
-                if (Thread::IsRunning() == true) {
-                    _continue = true;
-                    result = Core::ERROR_INPROGRESS;
-                } else {
+                bool expected = false;
+
+                // Attempt to set the `_continue` flag to `true` and start the thread.
+                if (_continue.compare_exchange_strong(expected, true, std::memory_order_acq_rel)) {
```
pwielders, line 946: Thread class is by definition data safe. Calling Run() twice is not an issue, so why protect with an atomic? To get a proper response error return? Are you using the return value? I checked and I think it is only used in the Render() who does not use the outcome of the trigger at all. Suggest to simplify.. Just call run and return *no* result...

pwielders (approved):

PR #897, opened 2025-04-28
[RemoteControl] fix keyboards and mouses detection

pgorszkowski-igalia:
In case of keyboards or mouses which names does not contains words "keyboard" or "mouse" they are not recognized as input devices. To solve it we need to iterate through available devices with udev and add them based on properties like: ID_INPUT_KEYBOARD or ID_INPUT_MOUSE.


pwielders:
Nice contribution. Thanks @pgorszkowski-igalia !

PR #898, opened 2025-05-05
Test plugin to verify the names of the JSON-RPC functions

nxtum:


VeithMetro (approved):
Looks good!

sebaszm (changes requested):
I think the dummy methods should at least:
- TRACE the data so the tester knows it worked, and the parameters were passed.
- Then send them back via the notification (the interface needs to be modified to pass parameters in the event, currently it does not validate case convention in the events).

VeithMetro (approved):

PR #902, opened 2025-05-19
[SubsystemController] Update after changes to the interface

VeithMetro:
First, we have to merge [this](https://github.com/rdkcentral/ThunderInterfaces/pull/441) PR with changes to the interface

I removed the notification for now, because it was never exposed on JSON-RPC to begin with, as it wasn't present in the .json file. And on top of that, with the Deactivate removed, as it is no longer allowed to "manually" disable subsystems, I suppose a notification is not very useful to have... but I can bring it back and expose on the interface, if we see it still has any value


sebaszm (approved):
The notification from IController::ISubsystems is supperior anywyay

PR #924, opened 2025-09-26
Development/compositor enhancements

bramoosterhuis:
- The compositor now also runs on **NVIDIA hardware** with the Wayland backend.  
- Buffer handling has been reworked:  
  - The **client is now fully responsible** for buffer allocation, pixel format, and modifier selection.  
  - The compositor only allocates a **Client slot**. Once the client is ready, it offers a **DMA-BUF** back to the compositor, which is then used to create a texture for rendering on the output buffers.  
- Clients can now **offer multiple buffers (up to 4)** simultaneously.  
- To support this, `ServerBufferType` has been moved from the `Client` object to a new **Buffer** object.  
- The client must request rendering on the correct `ClientBufferType` within its own process space. The corresponding `ServerBufferType` is then selected as the active content buffer for rendering.  

related PR
https://github.com/rdkcentral/ThunderClientLibraries/pull/321


pwielders (changes requested):
Minor: Can you remove all cerr streams. They cluther up the CLI, if it is running!

pwielders (approved):

PR #926, opened 2025-10-03
Development/batch jsonrpc

bramoosterhuis:
Thunder plugin that enables batching multiple JSON-RPC requests into a single HTTP or WebSocket call, with configurable limits and timeout handling.

### DIFF
--- a/JsonRpcMuxer/tests/test_jsonrpcmuxer.py
+++ b/JsonRpcMuxer/tests/test_jsonrpcmuxer.py
@@ -499,7 +499,7 @@ class JsonRpcMuxerTester:
             print_fail(f"Expected {total_runs * self.config.max_batches} results, got {len(results)}")
             self.failed += 1
     
-    def test_http_invalid_methode(self):
+    def test_http_invalid_method(self):
         """Test HTTP interface with an invalid method"""
         print_test("HTTP: Invalid method (should reject)")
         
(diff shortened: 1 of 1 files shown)


pwielders (changes requested):
Looks ggoood but I think you are missing a cleanup procedure if the Plugin is being deactivated. Please keep track of issued batches and kill them (stop processing, report error) if the plugin is deactivated.

pwielders (changes requested):
1) Make sure that in the Deinitialize the wait fro batches is infinite and print if, for example after 5S, you still have batches that you print all ongoing batches (method to invoke)
2) Look at the combinations atomic and criticalsections. Only use 1 technique, referably the atomic if possible.

bramoosterhuis:
I moved to Core::Critical sections because its a simpler approach. Timing wise I see that there is only benefits for heavy parallel processing, that was something you didn't want to encourage using. 

📊 1. Sequential mode
Test | no-atomic throughput | atomic throughput | Relative gain | Notes
-- | -- | -- | -- | --
rapid_fire | 374–454 avg ≈ 416 | 398–451 avg ≈ 423 | +1.7 % | Essentially identical, maybe measurement noise
high_concurrency | 252–369 avg ≈ 307 | 240–349 avg ≈ 287 | -6.5 % | Atomic slightly slower (overhead > benefit)
sustained_load | 2.97–2.99 ≈ 2.98 | 2.97–2.98 ≈ 2.98 | 0 % | Identical, CPU idle
maxbatches_limit | 124–160 avg ≈ 138 | 121–132 avg ≈ 126 | -8.7 % | Slightly worse with atomic (extra overhead)

✅ Summary (Sequential):
Atomic gives no practical speedup, and sometimes small slowdowns.

📊 2. Parallel mode
Test | no-atomic throughput | atomic throughput | Relative gain | Notes
-- | -- | -- | -- | --
rapid_fire | 413–434 avg ≈ 427 | 424–503 avg ≈ 475 | +11.2 % | Atomic reduces contention, real measurable win
high_concurrency | 306–321 avg ≈ 314 | 316–490 avg ≈ 402 | +28 % | Clear improvement under concurrency
sustained_load | 2.96–2.98 ≈ 2.97 | 2.97–2.98 ≈ 2.98 | 0 % | Same, as expected
maxbatches_limit | 127–148 avg ≈ 134 | 116–220 avg ≈ 156 | +16 % | Nice gain at saturation point

✅ Summary (Parallel):
Atomic improves throughput by ~10–30 % when multiple threads push batches concurrently.
The benefit is consistent across bursty and sustained high-load tests,
but irrelevant under light load.

📈 Relative overall performance
Mode | Atomic gain (average) | Meaning
-- | -- | --
Sequential | ~ 0 % (±5 %) | Within noise margin; no advantage
Parallel | +15–20 % overall | Worth it only under heavy concurrency

🧠 Interpretation

- Atomic only helps when multiple worker threads compete for shared counters (as in parallel mode).
- Under UI-level throughput, that gain is irrelevant.
- In production or long-term maintenance, the debugging simplicity of locks wins.

PR #947, opened 2025-12-12
Adjust all places where RPC::IteratorType is created to use std::vector

VeithMetro:


sebaszm (changes requested):
In some cases we have an idea how much elements will be pushed to the vector, so I suppose reserve() would be beneficial?

sebaszm (dismissed):

sebaszm (approved):

PR #968, opened 2026-05-27
Adjust to new Common folder

nxtum:


sebaszm (changes requested):
Optimally the "common" folder solution must not require changes to existing plugins.

PR #972, opened 2026-06-22
Add TestStateControl plugin for IStateControl interface testing

smanes0213:
Introduce a minimal TestStateControl plugin that implements IPlugin and IStateControl.

State transitions are dispatched asynchronously via Core::WorkerPool::JobType. The plugin fires IStateControl::INotification callbacks on every real state change and emits EXITED during Deinitialize().

Also includes a GTest integration test suite (StateControlTestSuite) that uses ThunderTestRuntime to spin up a real in-process PluginHost::Server and exercise the full IStateControl contract:
  - State transitions (UNINITIALIZED, RESUMED, SUSPENDED)
  - Idempotent requests
  - Observer registration, notification, and unregistration
  - Framework auto-resume when Resumed=true in plugin config
  - IShell::Resumed() reflection
  - IStateControl reachability via QueryInterfaceByCallsign


tests/TestStateControl/tests/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+
+target_link_libraries(${TARGET}
+    PRIVATE
+        thunder_test_support::thunder_test_support
+        GTest::gtest
+        GTest::gtest_main
+)
+
+set_target_properties(${TARGET} PROPERTIES
+    CXX_STANDARD 17
```
bramoosterhuis: The rest of the code uses `${CXX_STD}` here. Any reason to hardcode `17` instead?

tests/TestStateControl/tests/StateControlTestSuite.cpp
```diff
@@ -0,0 +1,813 @@
...
+        cfg.Resumed   = false;
+
+        const uint32_t result = _runtime.Initialize({ cfg });
+        ASSERT_EQ(result, Core::ERROR_NONE)
+            << "ThunderTestRuntime failed to initialise (error " << result << ")";
+    }
+
+    static void TearDownTestSuite()
+    {
+        _runtime.Deinitialize();
```
bramoosterhuis, line 229: Nothing is failing right now, but to keep each fixture isolated and starting from a clean slate I'd add `Core::Singleton::Dispose()` after `_runtime.Deinitialize()` in each `TearDownTestSuite()`. The runtime tears down the server but not the process-wide singletons (e.g. `ResourceMonitor` and its thread), so right now they carry over between the four fixtures. Disposing here is safe since the server is already deleted.

bramoosterhuis (changes requested):

bramoosterhuis (dismissed):

sebaszm:
Please sync actions exectuion with upcoming VolumeControl tests, consult with @VeithMetro.

VeithMetro (changes requested):
Let's adjust the Actions as discussed on Teams 😄

.github/workflows/Linux build template.yml
```diff
@@ -100,7 +100,8 @@ jobs:
     - name: Set test configuration
       if: matrix.architecture == 'tests64'
       run: |
-        echo "TESTS_CMAKE_FLAGS=-DVOLUMECONTROL_TESTS=ON" >> "$GITHUB_ENV"
+        echo "TESTS_CMAKE_FLAGS=$TESTS_CMAKE_FLAGS -DVOLUMECONTROL_TESTS=ON" >> "$GITHUB_ENV"
+        echo "TESTS_CMAKE_FLAGS=$TESTS_CMAKE_FLAGS -DTESTSTATECONTROL_TESTS=ON" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
VeithMetro: @smanes0213 Ah, indeed Copilot is correct, I was not aware that GitHub variables are updated only once, at the end of each step... In that case, I propose we just change this step to something like this if you agree:
```
# Add "-DFOO_TESTS=ON" to the FLAGS line below for each new test module
    - name: Set test configuration
      if: matrix.architecture == 'tests64'
      run: |
        FLAGS="-DVOLUMECONTROL_TESTS=ON -DTESTSTATECONTROL_TESTS=ON"
        echo "TESTS_CMAKE_FLAGS=$FLAGS" >> "$GITHUB_ENV"
```
We could even make it just one line, and write to `TESTS_CMAKE_FLAGS` right away, but I think it is more readable as two separate lines. In the end, I think this is even cleaner than before 😄

.github/workflows/Build ThunderNanoServices on Windows.yml
```diff
@@ -11,6 +11,7 @@ on:
     branches: ["master"]
 
 env:
+  common: Thunder\Source\common\common.vcxproj
```
VeithMetro: Let's remove any changes in this workflow. The fact that it is failing is not connected to this PR, so let's not try to fix it here.

Also, we discussed with @nxtum on how to fix this, since he introduced the common lib, and we agreed that the best approach is simply to add the common project as a dependancy for all the plugin 😄

.github/workflows/Linux build template.yml
```diff
@@ -46,7 +46,7 @@ jobs:
           PKGS="$PKGS zlib1g-dev:i386 libssl-dev:i386 libsbc-dev:i386 gcc-13-multilib g++-13-multilib"
           PKGS="$PKGS libicu-dev:i386 libpng-dev:i386 libdrm-dev:i386 libgbm-dev:i386 libegl-dev:i386 libgles2-mesa-dev:i386"
         else
-          PKGS="$PKGS zlib1g-dev libssl-dev libsbc-dev"
+          PKGS="$PKGS zlib1g-dev libssl-dev libsbc-dev libgtest-dev"
```
VeithMetro: I believe this should no longer be needed

.github/workflows/Linux build template.yml
```diff
@@ -95,12 +95,11 @@ jobs:
...
         search_string: ${{github.event.pull_request.body}}
 
-# Add one line per new test module:
-# echo "TESTS_CMAKE_FLAGS=$TESTS_CMAKE_FLAGS -DFOO_TESTS=ON" >> "$GITHUB_ENV"
+# Add one flag per new test module to the TESTS_CMAKE_FLAGS variable below:
     - name: Set test configuration
       if: matrix.architecture == 'tests64'
       run: |
-        echo "TESTS_CMAKE_FLAGS=-DVOLUMECONTROL_TESTS=ON" >> "$GITHUB_ENV"
+        echo "TESTS_CMAKE_FLAGS=$TESTS_CMAKE_FLAGS -DVOLUMECONTROL_TESTS=ON -DTESTSTATECONTROL_TESTS=ON" >> "$GITHUB_ENV"
```
VeithMetro: I think there is no need to append this variable, as it should be empty before this point:
```
echo "TESTS_CMAKE_FLAGS=-DVOLUMECONTROL_TESTS=ON -DTESTSTATECONTROL_TESTS=ON" >> "$GITHUB_ENV"
```

VeithMetro (changes requested):

VeithMetro (dismissed):
Actions look correct now, good job! 😄

sebaszm (dismissed):

sebaszm (approved):

PR #976, opened 2026-07-15
Optimizing the Benchmark QA plugin

smanes0213:
- Shell (Benchmark) implements IBenchmark directly; JSON-RPC via JBenchmark::Register(*this, this) — removes BenchmarkNotification, RegisterJsonRpcHandlers, UnregisterJsonRpcHandlers 
- Add IBenchmark method implementations: Trigger, CollectData, LatencyThreshold, MemoryThreshold, Register, Unregister 
- LatencyThreshold/MemoryThreshold accept object params {"value": N} aligned with generated JsonData::Benchmark types
- Replace Core::StopWatch (µs resolution) with clock_gettime(CLOCK_MONOTONIC) for true nanosecond timing 
- Drop batch-derived avgNs override; all stats from same per-call source — fixes avgNs > maxNs on fast in-process calls


tests/Benchmark/BenchmarkImplementation.cpp
```diff
@@ -83,6 +83,12 @@ namespace Plugin {
             return Core::ERROR_NONE;
         }
 
+        Core::hresult SetBaseline(IBenchmarkResultIterator* /* baseline */) override
+        {
+            // Baseline management is handled on the plugin shell side.
+            return Core::ERROR_NONE;
+        }
```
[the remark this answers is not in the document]
bramoosterhuis: `ERROR_NONE` here is misleading: a caller that reaches this layer directly will think the baseline was set, while nothing happens. It's technically more correct to use `ERROR_NOT_SUPPORTED`, or `ASSERT(false && "Should not be called here");`/`ASSERT(false);` if you want to enforce that this path is never taken.

tests/Benchmark/Benchmark.cpp
```diff
@@ -289,68 +419,128 @@ namespace Plugin {
...
+            return Core::ERROR_BAD_REQUEST;
+        }
+        // Reject concurrent triggers: interleaved results would corrupt the result set
+        // and produce an incorrect allPassed evaluation.
+        if (_triggerRunning.test_and_set()) {
+            return Core::ERROR_INPROGRESS;
+        }
+        TriggerGuard guard(_triggerRunning);
+        _adminLock.Lock();
+        _results.clear();
```
bramoosterhuis: `_results.clear()` here, then `RunPayloadBenchmarks()` fills `_results` incrementally (one lock per `addResult`, not one lock over the full run). A concurrent `CollectData()` can land mid-run and return a partial result set (e.g. 5 of 12 methods) with no signal that it's incomplete. Narrow race window (needs a second client calling `CollectData()` while another client's `Trigger()` is running), but silent failure mode.

Suggest `RunPayloadBenchmarks()` builds into a local vector and swaps into `_results` under a single lock once the full run is done, instead of clearing here and filling incrementally:

``` c++
std::vector<QualityAssurance::IBenchmark::BenchmarkResult> local;
auto addResult = [&local](QualityAssurance::IBenchmark::BenchmarkResult&& r) {
    local.push_back(std::move(r));
};
... (all addResult calls as before) ...
_adminLock.Lock();
_results = std::move(local);
_adminLock.Unlock();
```

That also removes the need for this `clear()` at the top of `Trigger()`, the swap replaces the whole vector atomically.

bramoosterhuis (changes requested):

tests/Benchmark/Benchmark.cpp
```diff
@@ -289,78 +419,134 @@ namespace Plugin {
...
     {
-        PluginHost::JSONRPC::Unregister(_T("Trigger"));
-        PluginHost::JSONRPC::Unregister(_T("CollectData"));
-        PluginHost::JSONRPC::Unregister(_T("LatencyThreshold"));
-        PluginHost::JSONRPC::Unregister(_T("MemoryThreshold"));
+        if (sink == nullptr) return Core::ERROR_BAD_REQUEST;
+        _adminLock.Lock();
+        auto it = std::find(_notifications.begin(), _notifications.end(), sink);
+        if (it != _notifications.end()) {
+            (*it)->Release();
```
bramoosterhuis: I missed this one earlier, but Release() is called while _adminLock is held, and the same applies to
AddRef() in Register() and the loop in Deinitialize(). Trigger() (lines
437-447) does it the other way round: copy the vector and AddRef under the
lock, unlock, then notify and Release outside.

Since _notifications is only populated by external clients over COM-RPC, the
sinks here are proxies, so Release() marshals a RELEASE message and touches
RPC::Administrator state (IUnknown.cpp:95). That means _adminLock is held
across an IPC round trip, blocking any concurrent CollectData(), Trigger() or
threshold setter for its duration, and it creates a lock-ordering dependency
between _adminLock and the COM-RPC administrator locks that Trigger()
deliberately avoids.

Suggest matching Trigger():

    _adminLock.Lock();
    auto it = std::find(_notifications.begin(), _notifications.end(), sink);
    QualityAssurance::IBenchmark::INotification* toRelease = nullptr;
    if (it != _notifications.end()) {
        toRelease = *it;
        _notifications.erase(it);
    }
    _adminLock.Unlock();

    if (toRelease != nullptr) {
        toRelease->Release();
    }

tests/Benchmark/Benchmark.cpp
```diff
@@ -248,9 +372,15 @@ namespace Plugin {
...
 
         if (_benchmark != nullptr) {
-            _benchmark->Unregister(&_benchmarkNotification);
+            QualityAssurance::JBenchmark::Unregister(*this);
+            PluginHost::JSONRPC::Unregister(_T("setBaseline"));
 
-            UnregisterJsonRpcHandlers();
+            _adminLock.Lock();
+            for (auto* sink : _notifications) {
+                sink->Release();
```
bramoosterhuis, line 295: Same as my comment on line 536: Release() runs under _adminLock, and these
sinks are COM-RPC proxies, so each Release() marshals a RELEASE message and
touches RPC::Administrator state. Here it's a loop, so the lock is held
across one IPC round trip per registered sink.

Swapping the vector out first releases the lock immediately and matches what
Trigger() already does:

    _adminLock.Lock();
    std::vector<QualityAssurance::IBenchmark::INotification*> sinks;
    sinks.swap(_notifications);
    _adminLock.Unlock();

    for (auto* sink : sinks) {
        sink->Release();
    }

After the swap _notifications is already empty, so the clear() isn't needed
any more.

bramoosterhuis (changes requested):

bramoosterhuis (dismissed):

sebaszm (approved):

